PR #158, opened 2023-06-21
feat: some small cleanups

marktoda:
- add natspec for some errors
- add unchecked in exclusivity lib
- some comments


Elenore23 (commented):
ok

PR #187, opened 2023-08-18
feat: add SwapRouter02ExecutorWithPermit to support 2612 style permit tokens

zhongeric:
Add SwapRouter02ExecutorWithPermit, an extension of the current sample executor which introduces two new entrypoints for filing orders: `executeWithPermit` and `executeBatchWithPermit`.

This allows fillers to fill orders sent by users who have not yet made the necessary approvals to Permit2 for the input tokens in their order. The executor will simply confirm the 2612 permit before calling the respective execute function on the reactor.


src/sample-executors/SwapRouter02Executor.sol
```diff
@@ -21,10 +21,10 @@ contract SwapRouter02Executor is IReactorCallback, Owned {
...
     error MsgSenderNotReactor();
 
-    ISwapRouter02 private immutable swapRouter02;
-    address private immutable whitelistedCaller;
-    IReactor private immutable reactor;
-    WETH private immutable weth;
+    ISwapRouter02 public immutable swapRouter02;
+    address public immutable whitelistedCaller;
+    IReactor public immutable reactor;
+    WETH public immutable weth;
```
zhongeric, line 27: making these public so swapRouter02ExecutorWithPermit can inherit

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
```diff
@@ -0,0 +1,59 @@
...
+contract SwapRouter02ExecutorWithPermit is SwapRouter02Executor {
+    constructor(address _whitelistedCaller, IReactor _reactor, address _owner, ISwapRouter02 _swapRouter02)
+        SwapRouter02Executor(_whitelistedCaller, _reactor, _owner, _swapRouter02)
+    {}
+
+    /// @notice assume that we already have all output tokens
+    /// @notice the reactor performs no verification that the user's signed permit is executed correctly
+    ///         e.g. if the necessary approvals are already set, a filler can call this function or the standard execute function to fill the order
+    /// @dev assume 2612 permit is collected offchain
+    function executeWithPermit(SignedOrder calldata order, bytes calldata callbackData, bytes calldata permitData)
```
marktoda, line 25: honestly seems like useful enough feature to just be in the main executor, like i dont see a reason someone would

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
```diff
@@ -0,0 +1,59 @@
...
+        _permitBatch(permitData);
+        reactor.executeBatchWithCallback(orders, callbackData);
+    }
+
+    /// @notice execute a signed 2612-style permit
+    /// the transaction will revert if the permit cannot be executed
+    /// must be called before the call to the reactor
+    function _permit(bytes calldata permitData) internal {
+        (address token, bytes memory data) = abi.decode(permitData, (address, bytes));
+        (address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
```
marktoda, line 49: why take permitData as bytes and abi decode it? would prefer just taking a struct

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
```diff
@@ -0,0 +1,59 @@
...
+        _permit(permitData);
+        reactor.executeWithCallback(order, callbackData);
+    }
+
+    /// @notice assume that we already have all output tokens
+    /// @dev assume 2612 permits are collected offchain
+    function executeBatchWithPermit(
+        SignedOrder[] calldata orders,
+        bytes calldata callbackData,
+        bytes[] calldata permitData
```
marktoda, line 38: doing array of permitData for executeBatch and single for execute single doesn't seem like a perfect matchup - like in the batch of orders they might not all need permits, so just cuz its a list of orders doesnt mean its necessarily a list of permits.. likewise maybe for executesingle we might want to send multiple permits for some reason?

i think we should just decouple it and have a multicall and permit function so the permits can be layered on granularly

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
```diff
@@ -0,0 +1,59 @@
...
+    /// must be called before the call to the reactor
+    function _permit(bytes calldata permitData) internal {
+        (address token, bytes memory data) = abi.decode(permitData, (address, bytes));
+        (address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
+            abi.decode(data, (address, address, uint256, uint256, uint8, bytes32, bytes32));
+        ERC20(token).permit(owner, spender, value, deadline, v, r, s);
+    }
+
+    function _permitBatch(bytes[] calldata permitData) internal {
+        for (uint256 i = 0; i < permitData.length; i++) {
```
marktoda, line 55: unchecked incrementor

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
marktoda: maybe we should make a base contract that other executors can inherit from that has this shared logic? wdyt

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
zhongeric: yeah i'm down to put in the original executor, was thinking that it could be nice to keep the old one as is but its mostly for reference / our own use

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
zhongeric: ooo good point about the multiple orders not all needing permits, it would be weird to have the inputs be different lengths...

how would you layer permits on granularly? are you thinking of bundling the orders together with any applicable permits like [ (order, permits[]), (order, permits[]), ...] ?

src/sample-executors/SwapRouter02ExecutorWithPermit.sol
zhongeric: `_permit(PermitData memory data)` ? and 
```
{
address token;
address owner;
...
}
```
I like that

zhongeric: for some reason i thought passing the calldata would be cheaper than a struct?

zhongeric: Taking as a struct saves gas :)

zhongeric:
Closing this in favor of #202

PR #202, opened 2023-09-12
refactor: add BaseExecutor with built in permit

zhongeric:
SwapRouter02Executor now inherits BaseExecutor which adds a new entrypoint `multicall` which allows callers to chain together series of execute and permit calls.


src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,110 @@
...
+
+import {Owned} from "solmate/src/auth/Owned.sol";
+import {SafeTransferLib} from "solmate/src/utils/SafeTransferLib.sol";
+import {ERC20} from "solmate/src/tokens/ERC20.sol";
+
+import {ResolvedOrder, SignedOrder} from "../base/ReactorStructs.sol";
+import {IReactorCallback} from "../interfaces/IReactorCallback.sol";
+import {IReactor} from "../interfaces/IReactor.sol";
+
+library Commands {
```
marktoda, line 12: ohhh interesting, reason why you decided to go with UR style commands vs a simple self delegatecall multisig? like https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/Multicall.sol

worried constructing these commands may be a pain for backend code compared to just encoding array of calldata

marktoda: I guess some gas benefit this way?

marktoda: Also what if inheritors of this base want to add more commands? they can't edit the base commands type. If using a simple multicall they can just add new external fns which can be multicalled...

IMO this feels a bit too hardcore for a sample executor, would prefer a simple multicall

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,110 @@
...
+
+    IReactor public immutable reactor;
+
+    constructor(IReactor _reactor, address _owner) Owned(_owner) {
+        reactor = _reactor;
+    }
+
+    function reactorCallback(ResolvedOrder[] calldata, bytes calldata callbackData) external virtual;
+
+    function _restrictCall() internal virtual {}
```
marktoda, line 40: comment on what htis is meant to be used for? assuming its auth, in which case seems like a strange thing to have as a top level virtual function. i.e. maybe some executors want different auth for different commands and so on? this model makes that impossible

src/sample-executors/BaseExecutor.sol
marktoda: Can check out Sara's old PR here which does this https://github.com/Uniswap/UniswapX/pull/115/files

marktoda (changes requested):

.forge-snapshots/SwapRouter02ExecutorExecuteAlreadyApproved.snap
```diff
@@ -1 +1 @@
-118092
\ No newline at end of file
+109807
```
snreynolds, line 1: Kinda confused why the gas is so high...Could you compare the snapshot taking out signOrder and the abi.encode stuff in the PR before? Or put it inside this snapshot. I think there will be slightly more accurate gas comparison that way

src/sample-executors/BaseExecutor.sol
snreynolds: Agree that a generic multicall would be a much easier lift

src/sample-executors/BaseExecutor.sol
snreynolds: er why not have it just live as a modifier? this looks like its just a wrapper over onlyWhitelistedCaller

zhongeric: Ah yeah will just do super.multicall to apply the whitelistedCaller modifier in SwapRouter02

.forge-snapshots/SwapRouter02ExecutorExecuteAlreadyApproved.snap
zhongeric: Ah yeah the gas savings are because it's not doing signOrder and encoding :( , gas should be relatively similar then

src/sample-executors/Multicall.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.16;
+
+/// @title Multicall
+/// @notice Code from https://github.com/Uniswap/v3-periphery/blob/main/contracts/base/Multicall.sol
+/// @notice Enables calling multiple methods in a single call to the contract
+abstract contract Multicall {
```
rileydcampbell, line 7: Can you just import this as an interface or external contract?

zhongeric: its from v3-periphery but that solidity version is different that this repo so can't directly import afaik

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    function executeBatch(SignedOrder[] memory orders, bytes memory callbackData) external payable {
+        reactor.executeBatchWithCallback{value: msg.value}(orders, callbackData);
+    }
+
+    /// @notice execute a signed 2612-style permit
+    /// the transaction will revert if the permit cannot be executed
+    function permit(PermitData memory permitData) public {
+        (address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
+            abi.decode(permitData.data, (address, address, uint256, uint256, uint8, bytes32, bytes32));
```
marktoda, line 39: im confused, why abi encode/decode all this instead of just have them as struct fields?

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,55 @@
...
+}
+
+abstract contract BaseExecutor is IReactorCallback, Multicall, Owned {
+    IReactor public immutable reactor;
+
+    constructor(IReactor _reactor, address _owner) Owned(_owner) {
+        reactor = _reactor;
+    }
+
+    function reactorCallback(ResolvedOrder[] calldata, bytes calldata callbackData) external virtual;
```
marktoda, line 25: natspecs?

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    function execute(SignedOrder memory order, bytes memory callbackData) external payable {
```
marktoda, line 27: should these be onlyOwner?

src/sample-executors/BaseExecutor.sol
marktoda: also should this be onlyReactor?

zhongeric: i dont think we can put the modifier on a virtual func

src/sample-executors/BaseExecutor.sol
zhongeric: since we also need the `token` address too, I thought it would make more sense to have 
```
struct {
token
data (all permit stuff here)
}
```
But yeah could just put everything top level in the struct

src/sample-executors/BaseExecutor.sol
zhongeric: mmm i thought we wanted execute to be gated by onlyWhitelstedCaller?

src/sample-executors/BaseExecutor.sol
snreynolds: ah yeah you can't. would add @dev comment in the `IReactorCallback` interface

zhongeric: sg, also added inheritdoc

src/sample-executors/BaseExecutor.sol
zhongeric: ugh also was notified that DAI's permit doesn't follow 2612

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,56 @@
...
+    IReactor public immutable reactor;
+
+    constructor(IReactor _reactor, address _owner) Owned(_owner) {
+        reactor = _reactor;
+    }
+
+    /// @inheritdoc IReactorCallback
+    function reactorCallback(ResolvedOrder[] calldata, bytes calldata callbackData) external virtual;
+
+    function execute(SignedOrder memory order, bytes memory callbackData) public payable virtual {
```
snreynolds, line 28: why is this virtual?

src/sample-executors/BaseExecutor.sol
snreynolds: yeah it doesn't :/ 

you could use Permit2Lib it handles special casing for DAI

snreynolds: I'd probably do that and then just put all the params in the PermitData struct

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,56 @@
...
+    function permit(PermitData memory permitData) public {
+        (address owner, address spender, uint256 value, uint256 deadline, uint8 v, bytes32 r, bytes32 s) =
+            abi.decode(permitData.data, (address, address, uint256, uint256, uint8, bytes32, bytes32));
+        ERC20(permitData.token).permit(owner, spender, value, deadline, v, r, s);
+    }
+
+    /// @notice execute a batch of signed 2612-style permits
+    /// the transaction will revert if any of the permits cannot be executed
+    function permitBatch(PermitData[] memory permitData) external {
+        for (uint256 i = 0; i < permitData.length;) {
```
snreynolds, line 47: save permitData.length outside of the for loop since your accessing it each time in the loop

snreynolds (changes requested):

src/sample-executors/BaseExecutor.sol
zhongeric: made this virtual so we can add the `onlyWhitelistedCaller` modifier to it

src/sample-executors/BaseExecutor.sol
zhongeric: Added in latest commit

src/sample-executors/BaseExecutor.sol
```diff
@@ -0,0 +1,56 @@
...
+import {IReactorCallback} from "../interfaces/IReactorCallback.sol";
+import {IReactor} from "../interfaces/IReactor.sol";
+import {Multicall} from "./Multicall.sol";
+
+struct PermitData {
+    address token;
+    bytes data;
+}
+
+abstract contract BaseExecutor is IReactorCallback, Multicall, Owned {
```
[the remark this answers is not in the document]
zhongeric: I dont think this contract is permit specific - adding the permit funcs available for multicall doesn't really add any overhead to the execution flow if you don't want to use them right?

src/sample-executors/BaseExecutor.sol
zhongeric: not a huge fan of this auth scheme, especially for `reactorCallback` because that **must** be restricted to only being called by the reactor but since its virtual in this base executor we can't enforce that here. So it's possible that integrators could make the mistake of not gating it by caller

src/sample-executors/BaseExecutor.sol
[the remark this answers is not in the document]
snreynolds: But now every integrator that just cares about the reactorCallback/execution inherits unused code if they don't care about permits?

src/sample-executors/BaseExecutor.sol
snreynolds: For this I'd include onlyReactor actually and just throw an error that its not implemented

snreynolds (changes requested):

src/sample-executors/BaseExecutor.sol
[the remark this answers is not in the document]
zhongeric: Renamed to BaseExecutorWithPermit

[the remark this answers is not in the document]
zhongeric: Actually, created BaseExecutor and BaseExecutorWithPermit so external integrators can just inherit BaseExecutor without the extra overhead from permit

marktoda:
closing as current preference is for fillers to do this in their fill contract

PR #204, opened 2023-09-26
chore: use Permit2 public repo in git submodules

rileydcampbell:


rileydcampbell:
> this looks like you deleted then readded permit2?

So one of our integrators was having an issue with using SSH to install these dependencies. Is there a reason we are using `git@github.com:Uniswap/permit2.git` instead of `https://github.com/Uniswap/permit2`?

snreynolds (approved):

snreynolds:
> > this looks like you deleted then readded permit2?
> 
> @snreynolds So one of our integrators was having an issue with using SSH to install these dependencies. Is there a reason we are using `git@github.com:Uniswap/permit2.git` instead of `https://github.com/Uniswap/permit2`?

gotcha, no thats fine, i think forge install doesnt use ssh so this is correct

PR #212, opened 2023-12-12
OZ Audit

marktoda:
This commit merges in all OZ audit commits from `development` into `main`. These were previously reviewed
https://github.com/Uniswap/UniswapX/pull/190
https://github.com/Uniswap/UniswapX/pull/189
https://github.com/Uniswap/UniswapX/pull/191
https://github.com/Uniswap/UniswapX/pull/193
https://github.com/Uniswap/UniswapX/pull/192
https://github.com/Uniswap/UniswapX/pull/194
https://github.com/Uniswap/UniswapX/pull/196
https://github.com/Uniswap/UniswapX/pull/197
https://github.com/Uniswap/UniswapX/pull/199


src/lib/DutchDecayLib.sol
```diff
@@ -28,7 +28,7 @@ library DutchDecayLib {
         view
         returns (uint256 decayedAmount)
     {
-        if (decayEndTime < decayStartTime) {
+        if (decayEndTime <= decayStartTime) {
```
hensha256, line 31: we really wanna revert if theyre the same? is it because they shouldnt be using a decay and should just have a limit order?

marktoda: https://github.com/Uniswap/UniswapX/pull/194

I think it was just undefined behavior what we should do in that case

marktoda: https://github.com/Uniswap/UniswapX/blob/main/audit/v1.1/OpenZeppelin.pdf L-03

hensha256 (approved):

PR #213, opened 2023-12-14
add uniswapx v2 initial structs

marktoda:
Add Initial UniswapXV2 reactor implementation


src/lib/V2DutchOrderLib.sol
```diff
@@ -0,0 +1,88 @@
...
+/// @notice helpers for handling v2 dutch order objects
+library V2DutchOrderLib {
+    using DutchOrderLib for DutchOutput[];
+    using OrderInfoLib for OrderInfo;
+
+    bytes internal constant V2_DUTCH_ORDER_TYPE = abi.encodePacked(
+        "V2DutchOrder(",
+        "OrderInfo info,",
+        "address cosigner,",
+        "address inputToken,",
```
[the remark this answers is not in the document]
hensha256: guess it saves gas

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,142 @@
...
+    /// - decayEndTime must be greater than or equal to decayStartTime
+    /// - if there's input decay, outputs must not decay
+    /// @dev Throws if the order is invalid
+    function _validateOrder(CosignedV2DutchOrder memory cosigned) internal pure {
+        V2DutchOrder memory order = cosigned.order;
+        if (order.inner.info.deadline < order.decayEndTime) {
+            revert DeadlineBeforeEndTime();
+        }
+
+        if (order.decayEndTime < order.decayStartTime) {
```
hensha256, line 111: should we revert for == too? thats what we just merged in the OZ audit fixes haha

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,142 @@
...
+            revert OrderEndTimeBeforeStartTime();
+        }
+
+        (bytes32 r, bytes32 s) = abi.decode(cosigned.signature, (bytes32, bytes32));
+        uint8 v = uint8(cosigned.signature[64]);
+        address signer = ecrecover(keccak256(abi.encode(order)), v, r, s);
```
hensha256, line 117: how come youre just abi encoding and then hashing and not using all the 712 stuff here?

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,142 @@
...
+        if (order.inner.cosigner != signer) {
```
hensha256, line 118: i think you need to check that cosigner isnt address(0) because ecrecover returns address(0) when it errors

hensha256 (commented):
some thoughts

src/reactors/V2DutchOrderReactor.sol
marktoda: oh because this will be signed by a server key, it doesn't need to see type info so we can save the extra hashing

src/lib/V2DutchOrderLib.sol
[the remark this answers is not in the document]
marktoda: yeah we flattened these for gas - this is just the way it parses in 712 views so no need for exact parity

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,142 @@
...
+        _updateWithOverrides(order);
+
+        resolvedOrder = ResolvedOrder({
+            info: order.inner.info,
+            input: order.inner.input.decay(order.decayStartTime, order.decayEndTime),
+            outputs: order.inner.outputs.decay(order.decayStartTime, order.decayEndTime),
+            sig: signedOrder.sig,
+            hash: order.hash()
+        });
+        resolvedOrder.handleOverride(order.exclusiveFiller, order.decayStartTime, order.exclusivityOverrideBps);
```
hensha256, line 75: maybe im missing something but i feel like the override logic is weird…
- first in validateOrder it validates the overrides exist and arent bigger/smaller than the original amount but not their %s
- then it goes through again and overrides the actual amounts of the order with the override values that were provided, without checking that they were big enough
- then it calls `handleOverride` which increases all of the values of the order with one increased by x%

hensha256: i feel like you definitely dont need both updateOverrides and handleOverrides

src/lib/V2DutchOrderLib.sol
```diff
@@ -0,0 +1,88 @@
...
+    // The time at which price becomes static
+    uint256 decayEndTime;
+    // The address who has exclusive rights to the order until decayStartTime
+    address exclusiveFiller;
+    // The amount in bps that a non-exclusive filler needs to improve the outputs by to be able to fill the order
+    uint256 exclusivityOverrideBps;
+    // The tokens that the swapper will provide when settling the order
+    uint256 inputOverride;
+    // The tokens that must be received to satisfy the order
+    uint256[] outputOverrides;
```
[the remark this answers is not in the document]
marktoda: inputOverride is for exactOutput trades, outputOverrides are for exactInput

> if a filler is overriding, theyre going to do so at exactly the bps that are required to override right (because why would they override by more).

note its not up to the filler to override in this case, these are strictly enforced on top of the inner order based on the secondary auction. True that fillers wouldn't do more than required, but that's ok in this case

src/reactors/V2DutchOrderReactor.sol
marktoda: ohh good catch

src/reactors/V2DutchOrderReactor.sol
marktoda: ohh ok theres 2 types of overrides - the first 2 are the newly introduced `cosigner` overrides which define the result of the secondary auction. Agreed we can remove the validation loop

the third bullet is the `filler override` which we already had in xv1, it basically is if some other public market filler that didn't compete in the auction has a significantly better price they can fill it at some fixed rate above the resolved price. I think we might want to remove that for v2 but keeping in anyways

marktoda: overloaded the terminology :D

marktoda: removing the public override for now

src/reactors/V2DutchOrderReactor.sol
marktoda: good call

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,133 @@
...
+            revert DeadlineBeforeEndTime();
+        }
+
+        if (order.cosignerData.decayEndTime <= order.cosignerData.decayStartTime) {
+            revert DutchDecayLib.EndTimeBeforeStartTime();
+        }
+
+        (bytes32 r, bytes32 s) = abi.decode(order.cosignature, (bytes32, bytes32));
+        uint8 v = uint8(order.cosignature[64]);
+        // cosigner signs over (orderHash || cosignerData)
```
ConjunctiveNormalForm, line 118: this means `orderHash` concat `cosignerData` right? at first I thought it's logical or

marktoda: oh yea its concat

hensha256: i was taught at university that this was the symbol for concat... but now that im googling i think that ive just been living a lie 😆 
image

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,133 @@
...
+        address signer = ecrecover(keccak256(abi.encodePacked(orderHash, abi.encode(order.cosignerData))), v, r, s);
```
hensha256, line 119: feels inefficient that `cosignature` is passed in as `(r || s || v)` and you decode that, just to then reencode it as `(v || r || s)` ? Can't people just pass it in in the right order?
And if you want to stick with this order, then I think you _should_ be able to move `v` without having to separate `r || s` and stick them back together... but thats harder and a nit pick lol

src/reactors/V2DutchOrderReactor.sol
```diff
@@ -0,0 +1,133 @@
...
+        V2DutchOrder memory order = abi.decode(signedOrder.order, (V2DutchOrder));
+        // hash the order _before_ overriding amounts, as this is the hash the user would have signed
+        bytes32 orderHash = order.hash();
+
+        _validateOrder(orderHash, order);
+        _updateWithOverrides(order);
+
+        resolvedOrder = ResolvedOrder({
+            info: order.info,
+            input: order.input.decay(order.cosignerData.decayStartTime, order.cosignerData.decayEndTime),
```
hensha256, line 60: Maybe I've misunderstood... but it feels like a strange UX that if a cosigner is overriding, they dont override with the actual amount that they want to fill at, but instead they override a new `startAmount` and have to predict what time their tx will be mined to get the correct `decay` to the amount they actually want to pay?
It would feel more natural to me that onchain it should calculate the `decay` and then call `_updateWithOverrides` above the _decayed_ value?

test/reactors/V2DutchOrderReactor.t.sol
```diff
@@ -0,0 +1,293 @@
...
+        uint256 overriddenOutputAmount = 1.1 ether;
+        uint256 inputAmount = 1 ether;
+        tokenIn.mint(swapper, inputAmount);
+        tokenOut.mint(address(fillContract), overriddenOutputAmount);
+        tokenIn.forceApprove(swapper, address(permit2), type(uint256).max);
+        CosignerData memory cosignerData = CosignerData({
+            decayStartTime: block.timestamp,
+            decayEndTime: block.timestamp + 100,
+            exclusiveFiller: address(0),
+            inputOverride: inputAmount,
```
hensha256, line 259: if youre not overriding the input for this test, shouldnt this just be 0?

test/reactors/V2DutchOrderReactor.t.sol
```diff
@@ -0,0 +1,293 @@
...
+        uint256 overriddenInputAmount = 0.7 ether;
+        tokenIn.mint(swapper, overriddenInputAmount);
+        tokenOut.mint(address(fillContract), outputAmount);
+        tokenIn.forceApprove(swapper, address(permit2), type(uint256).max);
+        CosignerData memory cosignerData = CosignerData({
+            decayStartTime: block.timestamp,
+            decayEndTime: block.timestamp + 100,
+            exclusiveFiller: address(0),
+            inputOverride: overriddenInputAmount,
+            outputOverrides: ArrayBuilder.fill(1, 1 ether)
```
hensha256, line 229: same here - shouldnt this just be 0s if youre not overriding?

hensha256 (dismissed):
left a couple of nits, but it looks great!

src/reactors/V2DutchOrderReactor.sol
marktoda: hmm thats true, ill try to pass them in correctly

marktoda: we could take them in already deconstructed i.e. a Signature struct with { v, r, s } but i think that ends up being more calldata

test/reactors/V2DutchOrderReactor.t.sol
marktoda: true, fixed

hensha256 (approved):

PR #215, opened 2023-12-19
feat: refactor dutch reactor tests

marktoda:
- Move common tests into a base test contract
- inherit it from Dutch and ExclusiveDutch (dutchv2 coming later)
- clean up test logic, use helpers and solarray

Solve the basic problem of reactors often having similar logic, and missing coverage due to assumptions that logic is the same so it should work the same. ExclusiveDutchReactor was missing a lot of coverage over the dutch validation for this reason


test/reactors/BaseDutchOrderReactor.t.sol
```diff
@@ -0,0 +1,431 @@
...
+            info: OrderInfoBuilder.init(address(reactor)).withDeadline(spec.deadline).withSwapper(address(swapper)),
+            decayStartTime: spec.startTime,
+            decayEndTime: spec.endTime,
+            input: spec.input,
+            outputs: spec.outputs
+        });
+        (order,) = createAndSignDutchOrder(request);
+    }
+
+    function test_dutch_resolveOutputNotStarted() public {
```
marktoda, line 78: fyi all these tests come from the validation section of `DutchOrderReactorTest`, but they've been cleaned up and refactored into this abstract test file which can be imported by any reactor test to add coverage to any dutch decaying order types

hensha256 (approved):

PR #217, opened 2024-01-21
delete the wrong description for execute and executeBatch in sample-executors /SwapRouter02Executor.sol?

0xkenta:
In the case of the Direct Filler, fillers need to own output tokens in advance. However, they don’t need to have them when they use their strategies in the filler contract. That’s why the description for both functions execute and executeBatch in the SwapRouter02Executor.sol file has been deleted in this pull request.


marktoda (dismissed):

marktoda:
ah need signed commits

0xkenta:
I submitted the new PR because of the signed commit. https://github.com/Uniswap/UniswapX/pull/224/files

PR #221, opened 2024-02-01
Packing overrides and filler

hensha256:
Very open to renaming things in this PR

extrabytes is used to encode any or all of exclusiveFiller, inputOverride, and outputOverrides.
If none of these 3 fields are needed for a tranasction, extraBytes is empty

In the case where at least 1 of these values is included:
First byte:
- bit 1: bool `hasExclusiveFiller`
- bit 2: bool `hasInputOverride` 
- bit3: bool `hasOutputOverrides`
- bits 4-8: `outputOverrides.length`

After this first byte, the relevant fields are appended in the order `exclusiveFiller`, `inputOverride`, `outputOverrides`.
`exclusiveFiller` is 20 bytes, `inputOverride` is 32 bytes, and `outputOverrides` is `(32 * outputOverrides.length)` bytes.

---------------------

An `exactInput` trade with an exclusive filler (assuming the filler overrides the outputs), should save:
- 32 bytes input override
- 32 bytes output overrides length
- 12 bytes of address passing
- and adds 1 byte
--> saving 75 x 0 bytes

This isn't _huge_ gas savings on mainnet (though is still decent as you can see) but is much higher savings on L2s


src/lib/CosignerDataLib.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+struct CosignerData {
+    // The time at which the DutchOutputs start decaying
+    uint256 decayStartTime;
+    // The time at which price becomes static
+    uint256 decayEndTime;
+    // Encoded exclusiveFiller, inputOverride, and outputOverrides, where needed
+    // First byte: fff0 0000 where f is a flag signalling inclusion of the 3 variables
```
marktoda, line 10: and the last 5 bits are the length of outputs array?

src/lib/CosignerDataLib.sol
```diff
@@ -0,0 +1,96 @@
...
+}
+
+library CosignerExtraDataLib {
+    bytes32 constant EXCL_FILLER_FLAG_MASK = 0x8000000000000000000000000000000000000000000000000000000000000000;
+    bytes32 constant INPUT_OVERRIDE_FLAG_MASK = 0x4000000000000000000000000000000000000000000000000000000000000000;
+    bytes32 constant OUTPUT_OVERRIDE_FLAG_MASK = 0x2000000000000000000000000000000000000000000000000000000000000000;
+    bytes32 constant OUTPUTS_LENGTH_FLAG_MASK = 0x1F00000000000000000000000000000000000000000000000000000000000000;
+
+    // "True" first bit of byte 1 signals that there is an exclusive filler
+    function hasExclusiveFiller(bytes memory extraData) internal pure returns (bool flag) {
```
marktoda, line 24: would it be even cheaper to just take a `bytes1 flags` or `bytes32 flags` here? so we only have to do the mload once? or is that canceled out by the masking and expansion?

src/lib/CosignerDataLib.sol
```diff
@@ -0,0 +1,96 @@
...
+            flag := and(mload(add(extraData, 32)), OUTPUT_OVERRIDE_FLAG_MASK)
+        }
+    }
+
+    function decodeExtraParameters(bytes memory extraData)
+        internal
+        pure
+        returns (address filler, uint256 inputOverride, uint256[] memory outputOverrides)
+    {
+        if (extraData.length == 0) return (filler, inputOverride, outputOverrides);
```
marktoda, line 52: can just do empty return right?

marktoda: also in what case would we ever have 0 length?

src/lib/CosignerDataLib.sol
```diff
@@ -0,0 +1,96 @@
...
+        // The first 32 bytes are length
+        // The first byte (index 0) only contains flags of whether each field is included in the bytes
+        // So we can start from index 1 (after 32) to start decoding each field
+        uint256 bytesOffset = 33;
+
+        if (hasExclusiveFiller(extraData)) {
+            // + 20 bytes for address, - 32 bytes for the length offset
+            require(extraData.length >= bytesOffset - 12);
```
marktoda, line 60: can we do a nicer custom err here?

src/lib/CosignerDataLib.sol
```diff
@@ -0,0 +1,96 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+struct CosignerData {
+    // The time at which the DutchOutputs start decaying
+    uint256 decayStartTime;
+    // The time at which price becomes static
+    uint256 decayEndTime;
```
marktoda, line 8: any benefit to including these in extraData as well and having them be only uint32s? that may also allow slightly cleaner reading as you could just have `parseInputOverrides(bytes encodedCosignerData) -> CosignerData` or something

test/reactors/V2DutchOrderReactor.t.sol
```diff
@@ -306,6 +280,33 @@ contract V2DutchOrderTest is PermitSignature, DeployPermit2, BaseDutchOrderReact
         fillContract.execute(signedOrder);
     }
 
+    function testDecodeExtraParametersFuzz(
+        address filler,
+        uint256 inputOverride,
+        uint256 outputOverride,
+        uint256 outputOverrideLength
```
marktoda, line 287: I think you can take arrays in the fuzzer if you want to test different amounts

marktoda (commented):
siick!!

src/lib/CosignerDataLib.sol
hensha256: length==0 when there is no exclusive filler, no inputOverrides, and no outputoverrides... so i guess an open order where someone fills the exact amount in the order itself

src/lib/CosignerDataLib.sol
hensha256: oh yeah good point i'll try that

marktoda:
skipping for now

PR #224, opened 2024-02-10
delete the wrong description for execute and executeBatch in sample-executors /SwapRouter02Executor.sol?

0xkenta:
In the case of the Direct Filler, fillers need to own output tokens in advance. However, they don’t need to have them when they use their strategies in the filler contract. That’s why the description for both functions execute and executeBatch in the SwapRouter02Executor.sol file has been deleted in this pull request.


0xkenta:
I had the approval for this PR from @marktoda. However, I can not submit the signed commit with the previous PR https://github.com/Uniswap/UniswapX/pull/217, so closed the last PR and this is the new one.

PR #226, opened 2024-02-13
feat: rename cosigner overrides

marktoda:
more clear name, and to avoid duplication with the exclusive filler
override. Now baseInput and baseOutputs are signed over by the swapper,
and overridden by inputAmount and outputAmounts by the cosigner


hensha256 (approved):
LOVE IT

PR #234, opened 2024-03-20
feat: add basic integ test

marktoda:
more cases to come, but this just ensures that orders created by the SDK
are actuallye executable by the reactor

Also fixes a couple bugs to make the above true


pannu5788:
Go Back to school

PR #235, opened 2024-03-28
chore: add orderQuoter as part of v2 deployment

ConjunctiveNormalForm:


marktoda:
we shouldn't need a separate orderquoter for V2 as quoter is reactor-agnostic

wonder if wouldnt make more sense to have a separate quoter-only script

PR #239, opened 2024-04-05
feat: fix comments

marktoda:
Issue: Cantina #16

### DIFF
--- a/src/reactors/V2DutchOrderReactor.sol
+++ b/src/reactors/V2DutchOrderReactor.sol
@@ -14,7 +14,7 @@ import {SignedOrder, ResolvedOrder} from "../base/ReactorStructs.sol";
 /// @dev resolution behavior:
 /// - If cosignature is invalid or not from specified cosigner, revert
 /// - If inputAmount is 0, then use baseInput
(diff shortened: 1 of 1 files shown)


src/reactors/V2DutchOrderReactor.sol
```diff
@@ -107,8 +107,8 @@ contract V2DutchOrderReactor is BaseReactor {
     }
 
     /// @notice validate the dutch order fields
-    /// - deadline must be greater than or equal than decayEndTime
-    /// - decayEndTime must be greater than or equal to decayStartTime
+    /// - deadline must be greater than or equal to decayEndTime
+    /// - decayEndTime must be greater than decayStartTime
```
zhongeric, line 111: so orders with no decay should have same start/end amounts but must still specify different decay start and end times?

zhongeric (approved):

src/reactors/V2DutchOrderReactor.sol
marktoda: yeah, it doesn't really matter what they are if start/end amounts are the same though

PR #241, opened 2024-04-05
feat: early return dutch decay if none

marktoda:
If startAmount == endAmount, early return. This is always true for
either the inputs or outputs of an order, so saves the decay calculation
half the time


src/lib/DutchDecayLib.sol
```diff
@@ -28,7 +28,9 @@ library DutchDecayLib {
         view
         returns (uint256 decayedAmount)
     {
-        if (decayEndTime <= decayStartTime) {
+        if (startAmount == endAmount) {
```
zhongeric, line 31: a little strange that the decay times are not validated if startAmount==endAmount

zhongeric: wdyt about putting this check after the time based ones?

zhongeric (approved):

src/lib/DutchDecayLib.sol
marktoda: part of the point of this issue is that the time-based ones are irrelevant if startamount == endamount, which is always true for one of inputs or outputs. timings will be validated for whichever one is decaying

zhongeric (dismissed):

zhongeric (approved):

PR #247, opened 2024-04-19
feat: reject both input and output fees

marktoda:
this avoids a double-fee case from protocol fee controller


src/base/ProtocolFees.sol
```diff
@@ -67,13 +71,17 @@ abstract contract ProtocolFees is Owned {
             for (uint256 j = 0; j < outputsLength; j++) {
                 OutputToken memory output = order.outputs[j];
                 if (output.token == feeOutput.token) {
+                    if (inputFeeTaken) revert InputAndOutputFees();
```
zhongeric, line 74: maybe im misreading but this code will never be true since inputFeeTaken is assigned after this?

marktoda: Oh, it's inside the outside loop. basically:

for every fee output added:
- is it a fee-on-output? (loop thru all outputs to check)
- is it a fee-on-input?

and this is just disallowing both, so placement inside the outer loop doesnt matter

zhongeric: ah I see

zhongeric (approved):

PR #252, opened 2024-06-25
feat: add priority tax reactor

zhongeric:
Introduces a new reactor type, PriorityOrderReactor, which scales the output amount required to fill an order by the priority fee set in the filling transaction


src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+    uint256 constant MPS = 1e10;
+
+    /// @notice returns a scaled input using the current priority fee and mpsPerPriorityFeeWei
+    /// @notice the amount is scaled down to favor the swapper
+    /// @param input the input to scale
+    /// @param priorityFee the current priority fee
+    /// @return a scaled input
+    function scale(PriorityInput memory input, uint256 priorityFee) internal pure returns (InputToken memory) {
+        if (priorityFee * input.mpsPerPriorityFeeWei >= MPS) {
+            return InputToken({token: input.token, amount: 0, maxAmount: 0});
```
marktoda, line 22: ok crazy but is there ever a case where the MEV is so high they'll pay YOU?

src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    /// @notice returns scaled outputs using the current priority fee and mpsPerPriorityFeeWei
+    function scale(PriorityOutput[] memory outputs, uint256 priorityFee)
+        internal
+        pure
+        returns (OutputToken[] memory result)
+    {
+        uint256 outputLength = outputs.length;
+        result = new OutputToken[](outputLength);
+        unchecked {
```
marktoda, line 49: not needed in 0.8.2x

src/reactors/PriorityOrderReactor.sol
```diff
@@ -0,0 +1,78 @@
...
+    function _resolve(SignedOrder calldata signedOrder)
+        internal
+        view
+        override
+        returns (ResolvedOrder memory resolvedOrder)
+    {
+        PriorityOrder memory priorityOrder = abi.decode(signedOrder.order, (PriorityOrder));
+        _validateOrder(priorityOrder);
+
+        uint256 priorityFee = tx.gasprice - block.basefee;
```
marktoda, line 34: one thing to make sure we consider is how different L2s handle these fields - I assume the default handling is the same as mainnet eip-1559 but potentially some handle differently due to separation of L1 data fee, slightly different gas rules etc.
Worth adding a comment to the contract I think clarifying that this only works on chains that have xyz properties

marktoda (commented):
very sick - few small thoughts

src/lib/PriorityFeeLib.sol
zhongeric: discussed offline - don't think this can happen since the MEV is naturally capped at the max input offered by the swapper

src/reactors/PriorityOrderReactor.sol
zhongeric: oooo good callout, so it's possible that this calculated value could include other additional fees / not represent the actual priority fee correctly?

src/lib/PriorityOrderLib.sol
```diff
@@ -0,0 +1,106 @@
...
+    function hash(PriorityOrder memory order) internal pure returns (bytes32) {
+        return keccak256(
+            abi.encode(
+                ORDER_TYPE_HASH,
+                order.info.hash(),
+                order.startBlock,
+                order.input.token,
+                order.input.amount,
+                order.input.mpsPerPriorityFeeWei,
+                hash(order.outputs)
```
snreynolds, line 102: order.outputs.hash() syntax?

src/reactors/PriorityOrderReactor.sol
```diff
@@ -0,0 +1,81 @@
...
+            revert InvalidDeadline();
+        }
+
+        if (order.startBlock > block.number) {
+            revert OrderNotFillable();
+        }
+
+        if (order.input.mpsPerPriorityFeeWei > 0) {
+            for (uint256 i = 0; i < order.outputs.length; i++) {
+                if (order.outputs[i].mpsPerPriorityFeeWei > 0) {
```
snreynolds, line 75: oh ok so you can only scale either the input or the output?

zhongeric: Yes only one can be scaled at a time

zhongeric: For exact in / exact output

src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    /// @notice we denominate priority fees in terms of milli-bips, or one thousandth of a basis point
+    uint256 constant MPS = 1e7;
+
+    /// @notice returns a scaled input using the current priority fee and mpsPerPriorityFeeWei
+    /// @notice this value is bounded by 0 since the maximal extractable value is <= the full input amount
+    /// @notice the amount is scaled down to favor the swapper
+    /// @notice maxAmount is set to be the original amount and is used to rebuild the permit2 token permissions struct
+    /// @param input the input to scale
+    /// @param priorityFee the current priority fee
```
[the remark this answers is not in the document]
zhongeric: priority fee should be in wei

src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+
+import {OutputToken, InputToken} from "../base/ReactorStructs.sol";
+import {PriorityOutput, PriorityInput} from "../lib/PriorityOrderLib.sol";
+import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
+
+/// @notice helpers for handling priority order objects
+library PriorityFeeLib {
+    using FixedPointMathLib for uint256;
+
+    /// @notice we denominate priority fees in terms of milli-bips, or one thousandth of a basis point
```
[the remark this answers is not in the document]
zhongeric: here's some reasoning:
- want to use relative scaling (increase by X bips per wei of priority fee)
- can't use bips as the denominator here because does not give fillers enough points to fill at on the curve (30 bps of price improvement = max bid of 30 wei)
- so we pick something reasonably smaller than bips but not too small where too much is being paid to the sequencer

src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+    /// @notice the amount is scaled down to favor the swapper
+    /// @notice maxAmount is set to be the original amount and is used to rebuild the permit2 token permissions struct
+    /// @param input the input to scale
+    /// @param priorityFee the current priority fee
+    /// @return a scaled input
+    function scale(PriorityInput memory input, uint256 priorityFee) internal pure returns (InputToken memory) {
+        if (priorityFee * input.mpsPerPriorityFeeWei >= MPS) {
+            return InputToken({token: input.token, amount: 0, maxAmount: input.amount});
+        }
+        uint256 scaledAmount = input.amount.mulDivDown((MPS - (priorityFee * input.mpsPerPriorityFeeWei)), MPS);
```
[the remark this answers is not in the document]
zhongeric: it's not guaranteed, mpsPerPriorityFeeWei * priorityFee is the number of mps to increase the outputs by. If that is >= MPS (total possible), then we set it equal to 0. This is the equivalent of saying that you want to reduce something by 100%.

src/lib/PriorityOrderLib.sol
```diff
@@ -0,0 +1,106 @@
...
+    address recipient;
+}
+
+/// @dev External struct used to specify priority orders
+struct PriorityOrder {
+    // generic order information
+    OrderInfo info;
+    // the block at which the order becomes active
+    uint256 startBlock;
+    // The tokens that the swapper will provide when settling the order
```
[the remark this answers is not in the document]
zhongeric: we need to put the scaling factor on both of them to denote if the order is exact in or exact out

src/lib/PriorityFeeLib.sol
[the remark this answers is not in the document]
snreynolds: Wouldnt that scenario cause an underflow here then? Cause we're doing 
`(MPS - (priorityFee * input.mpsPerPriorityFeeWei)`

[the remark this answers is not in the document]
snreynolds: oh i see its checked above! didn't catch that

src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+import {OutputToken, InputToken} from "../base/ReactorStructs.sol";
+import {PriorityOutput, PriorityInput} from "../lib/PriorityOrderLib.sol";
+import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
+
+/// @notice helpers for handling priority order objects
+library PriorityFeeLib {
+    using FixedPointMathLib for uint256;
+
+    /// @notice we denominate priority fees in terms of milli-bips, or one thousandth of a basis point
+    uint256 constant MPS = 1e7;
```
marktoda, line 13: Consider changing to pips after our chat w/ Dan?

src/lib/PriorityFeeLib.sol
```diff
@@ -0,0 +1,55 @@
...
+
+    /// @notice returns a scaled input using the current priority fee and mpsPerPriorityFeeWei
+    /// @notice this value is bounded by 0 since the maximal extractable value is <= the full input amount
+    /// @notice the amount is scaled down to favor the swapper
+    /// @notice maxAmount is set to be the original amount and is used to rebuild the permit2 token permissions struct
+    /// @param input the input to scale
+    /// @param priorityFee the current priority fee
+    /// @return a scaled input
+    function scale(PriorityInput memory input, uint256 priorityFee) internal pure returns (InputToken memory) {
+        if (priorityFee * input.mpsPerPriorityFeeWei >= MPS) {
```
marktoda, line 23: nit: cache `priorityfee * input.mpsPerPriorityFeeWei` for readability + maybe gas?

src/lib/PriorityOrderLib.sol
```diff
@@ -0,0 +1,106 @@
...
+    uint256 mpsPerPriorityFeeWei;
+    address recipient;
+}
+
+/// @dev External struct used to specify priority orders
+struct PriorityOrder {
+    // generic order information
+    OrderInfo info;
+    // the block at which the order becomes active
+    uint256 startBlock;
```
marktoda, line 28: I think this is good for now but one thing to consider for later --

this is signed over by the user, and users take unknown amt of sign to sign. In order to make sure this is in the future by the time the order is signed we will likely need to buffer quite a bit. May be difficult to tightly optimize for execution latency without some kind of cosigner situation just for updating the block

marktoda: potentially in the future users specify `minBlock` and `maxBlock` and labs can optionally cosign `targetBlock` with higher precision

test/lib/PriorityFeeLib.t.sol
```diff
@@ -0,0 +1,102 @@
...
+        PriorityOutput memory output =
+            PriorityOutput({token: address(0), amount: 1 ether, mpsPerPriorityFeeWei: 100, recipient: address(0)});
+
+        OutputToken memory scaledOutput = PriorityFeeLib.scale(output, tx.gasprice);
+        uint256 scaledAmount = (output.amount * (MPS + tx.gasprice * output.mpsPerPriorityFeeWei)) / MPS;
+
+        assertEq(scaledOutput.amount, scaledAmount);
+    }
+
+    function testScaleInputPriorityFee_fuzz(uint256 priorityFee) public {
```
marktoda, line 40: some non-fuzz edge case tests would be good:
- input fee over max defaults to 0
- super high / low priority fee

marktoda (commented):
Looks awesome!

src/lib/PriorityFeeLib.sol
marktoda: oh sara already said this

src/lib/PriorityOrderLib.sol
zhongeric: hmm this would require us to do `using PriorityOrderLib for PriorityOutput[]` within the same library? to get this syntax we would have to create a separate outputs library right

test/lib/PriorityFeeLib.t.sol
zhongeric: I added explicit tests for when the mpsPerPriorityFeeWei is 0, when the amount is large enough to trigger overflow in the math library, for small priority fees, and for when input fee defaults to 0

marktoda (dismissed):
LGTM!! :rocket:

marktoda (approved):

PR #254, opened 2024-07-09
feat: add cosigner logic and minPriorityFee

zhongeric:
image


src/lib/PriorityOrderLib.sol
```diff
@@ -24,18 +29,33 @@ struct PriorityOutput {
 struct PriorityOrder {
     // generic order information
     OrderInfo info;
+    // The address which must cosign the full order
+    address cosigner;
     // the block at which the order becomes active
-    uint256 startBlock;
+    uint256 auctionStartBlock;
```
marktoda, line 35: based on your diagram, the order is not active here yet but rather this is just the min that the cosigner can set it to right? maybe better name is `minTargetBlock`

src/lib/PriorityOrderLib.sol
```diff
@@ -24,18 +29,33 @@ struct PriorityOutput {
...
+    // the block at which the order can be executed without a cosigner
+    uint256 openAuctionStartBlock;
+    // the minimum priority fee required for filling the order
```
marktoda, line 38: unless I'm misunderstanding, this is more like the "baseline" priority fee above which fees are taxed, no?

src/lib/PriorityOrderLib.sol
```diff
@@ -24,18 +29,33 @@ struct PriorityOutput {
 struct PriorityOrder {
     // generic order information
     OrderInfo info;
+    // The address which must cosign the full order
+    address cosigner;
     // the block at which the order becomes active
-    uint256 startBlock;
+    uint256 auctionStartBlock;
+    // the block at which the order can be executed without a cosigner
+    uint256 openAuctionStartBlock;
```
marktoda, line 37: "open" is a little overloaded from v1/v2 wrt non-exclusivity - maybe "userAuctionStartBlock"?

marktoda: thinking about it more, do we actually need `auctionStartBlock` above? Seems what's happening here is:
- User sets an auction start block
- Cosigner can make it _earlier_ if they want
- is there even any risk of cosigner making it _too_ early? it doesn't really affect anything whether it's 1 second in the past or 1 year in the past afaict

src/reactors/PriorityOrderReactor.sol
```diff
@@ -16,31 +16,47 @@ contract PriorityOrderReactor is BaseReactor {
...
-        _validateOrder(priorityOrder);
+        PriorityOrder memory order = abi.decode(signedOrder.order, (PriorityOrder));
+        bytes32 orderHash = order.hash();
+
+        _updateWithCosignerData(order);
+        _validateOrder(orderHash, order);
 
         uint256 priorityFee = tx.gasprice - block.basefee;
+        if (priorityFee < order.minPriorityFeeWei) {
+            revert InsufficientPriorityFee();
```
marktoda, line 50: ah I wonder if these should still be allowed, and go through with 0 tax. I read this as "network is congested, so X prio fee is required to even get included. anything above that is MEV I want to tax". And then the filler pays <X, so either congestion went down or this order has very low MEV, so shouldn't really tax it

src/lib/PriorityOrderLib.sol
zhongeric: mmm so in practice, no risk of the cosigner making it too early. However an important note is that all of these values are in the future relative to the user signing the order, like auctionStartBlock is a few blocks in the future, preventing the cosigner from starting the order immediately.

src/reactors/PriorityOrderReactor.sol
zhongeric: ah yea I agree these should still go through

src/lib/PriorityOrderLib.sol
zhongeric: I like baseline

src/lib/PriorityOrderLib.sol
zhongeric: Ok I think I agree with you that we don't need auctionStartBlock. The ideal situation UX wise is to give the cosigner full control to set the target block whenever during the order lifecycle. However we want to bake in a period where the order can be submitted without the cosigner - so the question is where should this be? At the start of the order or at the end near the deadline?

Given that we are working with fuzzy numbers (buffer from frontend), I think its easiest to give ourselves more leeway here and have the "userAuctionStartBlock" be the parameter that is signed over by the user which is going to be pretty close to the deadline of the order. And then the cosigner can start the auction at any earlier block. As long as (userAuctionStartBlock - currentBlock) is enough time (doesn't have to be super long) this should be fine. 

The worry that the cosigner could start the order instantly is valid but something I think we just have to live with, it is not impossible to think that searchers could compete in < 2 seconds

ConjunctiveNormalForm: Huddled with @zhongeric on this exact topic too, and I think it comes down to whether or not we want to make orders fillable without cosigner data?

Just to recap how we got here:
- We are concerned that users can sit on the permit signature request for an unknown amount of time before signing, making the timestamp in the original request stale. This alone wouldn't break the system as we can just set deadline super loose.
  - However, we also want a small buffer in between when order is submitted, and when order becomes fillable, to allow order dissemination to enough fillers
- To combat this, we want to introduce a second timestamp which is later in time. We call it `auctionStartBlock`, and introduces a Lab-controlled cosigner that has the ability to set it to delay the auction start at the time of order submission
- However, introducing a cosigner means a new centralized entity is needed to run the auction, unless we introduce a third timestamp, `openAuctionStartBlock`, after which the cosigner is not required

I think I agree that we should design a system where Lab infra isn't  a necessary component, although I feel like we can set `openAuctionStartTime` tighter as opposed to be close to deadline to allow for faster fills in this edge case we are discussing

src/lib/PriorityOrderLib.sol
zhongeric: changed in latest commit

src/lib/PriorityOrderLib.sol
zhongeric: Re. our sync today:
- user signs on `auctionStartBlock`
- cosigner can override to decrease this by any amount
- order is valid _only_ after the resolved `auctionStartBlock`
- thus, order is valid after the min of (auctionTargetBlock, auctionStartBlock) until deadline

src/reactors/PriorityOrderReactor.sol
```diff
@@ -74,22 +73,17 @@ contract PriorityOrderReactor is BaseReactor {
 
     /// @notice update the order with the cosigner's data
     function _updateWithCosignerData(PriorityOrder memory order) internal view {
-        /// if there is a cosigned targetBlock, and we are not yet in the openAuction
         if (order.cosignerData.auctionTargetBlock != 0) {
```
marktoda, line 74: readability nit: early return if == 0

src/reactors/PriorityOrderReactor.sol
```diff
@@ -46,10 +44,11 @@ contract PriorityOrderReactor is BaseReactor {
         _validateOrder(orderHash, order);
 
         uint256 priorityFee = tx.gasprice - block.basefee;
-        if (priorityFee < order.minPriorityFeeWei) {
-            revert InsufficientPriorityFee();
+        if (priorityFee > order.baselinePriorityFeeWei) {
+            priorityFee -= order.baselinePriorityFeeWei;
```
marktoda, line 46: this can be unchecked

src/reactors/PriorityOrderReactor.sol
```diff
@@ -46,10 +44,11 @@ contract PriorityOrderReactor is BaseReactor {
         _validateOrder(orderHash, order);
 
         uint256 priorityFee = tx.gasprice - block.basefee;
```
marktoda, line 44: if tx.gasprice < block.basefee for some reason (maybe weird chain behavior etc) maybe we should either:
- revert with nicer message
- just use priority fee of 0 rather than revert?
Thoughts?

src/reactors/PriorityOrderReactor.sol
```diff
@@ -77,5 +101,13 @@ contract PriorityOrderReactor is BaseReactor {
                 }
             }
         }
+
+        (bytes32 r, bytes32 s) = abi.decode(order.cosignature, (bytes32, bytes32));
```
marktoda, line 105: don't you need to skip the cosigner check somehow if past the user's signed start block?

src/lib/PriorityOrderLib.sol
```diff
@@ -5,6 +5,11 @@ import {OrderInfo, InputToken, OutputToken} from "../base/ReactorStructs.sol";
 import {OrderInfoLib} from "./OrderInfoLib.sol";
 import {ERC20} from "solmate/src/tokens/ERC20.sol";
 
+struct PriorityCosignerData {
+    // the block at which the order can be executed (overrides auctionStartBlock)
+    uint256 auctionTargetBlock;
```
ConjunctiveNormalForm, line 10: can we flatten this, or do you think it's better to keep the same structure as v2 Order

zhongeric: I think better to keep as same structure since not much overhead for user as they are not signing over this

src/reactors/PriorityOrderReactor.sol
zhongeric: ah yes great point - I refactored a little bit to only do all of the cosigner logic if we are before the user's signed start block

src/reactors/PriorityOrderReactor.sol
zhongeric: hmmm so I thought a bit about this too,
- definitely agree with reverting with a custom error instead of the underflow
- I'm hesitant on doing priority fee of 0 because that basically skips the MEV tax mechanism for these orders, and I'm worried that a chain could change behavior and we wouldn't catch it because we have this silent fallback

zhongeric: ok synced - let's just revert with custom error

src/lib/PriorityOrderLib.sol
```diff
@@ -24,18 +29,31 @@ struct PriorityOutput {
 struct PriorityOrder {
     // generic order information
     OrderInfo info;
-    // the block at which the order becomes active
-    uint256 startBlock;
+    // The address which must cosign the full order
```
marktoda, line 32: nit: s/must/may?

src/reactors/PriorityOrderReactor.sol
```diff
@@ -56,17 +67,41 @@ contract PriorityOrderReactor is BaseReactor {
...
+    /// @param order the order to update
+    function _updateWithCosignerData(bytes32 orderHash, PriorityOrder memory order) internal pure {
+        /// return quickly if cosignerData is not set
+        if (order.cosignerData.auctionTargetBlock == 0) return;
+
+        if (order.cosignerData.auctionTargetBlock < order.auctionStartBlock) {
+            order.auctionStartBlock = order.cosignerData.auctionTargetBlock;
+        }
+        // validate cosigner signature
+        (bytes32 r, bytes32 s) = abi.decode(order.cosignature, (bytes32, bytes32));
```
marktoda, line 82: nit: may be nicer to have a small sig verification library now that we're doing this both here and in xv2

src/reactors/PriorityOrderReactor.sol
```diff
@@ -56,17 +67,41 @@ contract PriorityOrderReactor is BaseReactor {
...
+        uint8 v = uint8(order.cosignature[64]);
+        // cosigner signs over (orderHash || cosignerData)
+        address signer = ecrecover(keccak256(abi.encodePacked(orderHash, abi.encode(order.cosignerData))), v, r, s);
+        if (order.cosigner != signer || signer == address(0)) {
+            revert InvalidCosignature();
```
marktoda, line 87: What about if the user didn't specify a cosigner at all? Then this fails -- think we should allow that case right? Maybe w/ an early return above

src/reactors/PriorityOrderReactor.sol
```diff
@@ -56,17 +67,41 @@ contract PriorityOrderReactor is BaseReactor {
...
+    /// @notice update the priority order with the cosigner's data
+    /// @dev only called if the current block is before the auctionStartBlock signed by the user
+    /// @param orderHash the hash of the order
+    /// @param order the order to update
+    function _updateWithCosignerData(bytes32 orderHash, PriorityOrder memory order) internal pure {
+        /// return quickly if cosignerData is not set
+        if (order.cosignerData.auctionTargetBlock == 0) return;
+
+        if (order.cosignerData.auctionTargetBlock < order.auctionStartBlock) {
+            order.auctionStartBlock = order.cosignerData.auctionTargetBlock;
```
marktoda, line 79: Looking at this more, I wonder if we should just combine all of this with `validate` and if we actually even need to update the `auctionStartBlock` field in the order? It's only ever used for validation purposes afaict. 

```
...deadline check and stuff

uint256 auctionStartBlock = user's
if (not past the user's start block, and a cosigner is specified) check sig + update the local
check local against current, throw `OrderNotFillable`
```

marktoda (approved):
LGTM, feel free to merge and work on the rest in a followup

src/reactors/PriorityOrderReactor.sol
zhongeric: ah i thought that case would be caught by checking if the cosignerData auctionTargetBlock was the default val of 0

src/reactors/PriorityOrderReactor.sol
zhongeric: that is true! we don't really need to actually override the field ... on second look wait this might break P2 verification bc the order hash will change? and it's not caught by forge

marktoda: looks like you hash the order before overriding the field but yeah still cleaner to not update it imo

src/reactors/PriorityOrderReactor.sol
marktoda: oh, good point. Still adds a weird edge case where user didn't set a cosigner, but the filler sets a nonzero auctionTargetBloc and causes unnecessary failing

src/reactors/PriorityOrderReactor.sol
zhongeric: ah yea you're right

src/reactors/PriorityOrderReactor.sol
zhongeric: yea agree - changing it to check cosigner == address(0), more intuitive

PR #255, opened 2024-07-11
fix: cosigner verification edge cases and other fixes

zhongeric:


src/reactors/PriorityOrderReactor.sol
```diff
@@ -67,41 +63,33 @@ contract PriorityOrderReactor is BaseReactor {
...
 
+        uint256 auctionStartBlock = order.auctionStartBlock;
+
+        // we override auctionStartBlock with the cosigned auctionTargetBlock only if:
+        // - cosigner is specified
+        // - current block is before the auctionStartBlock signed by the user
+        // - cosigned auctionTargetBlock is before the auctionStartBlock signed by the user
+        if (
+            order.cosigner != address(0) && block.number < auctionStartBlock
+                && order.cosignerData.auctionTargetBlock < auctionStartBlock
```
marktoda, line 84: does it really matter if the cosigned auction block is past the user's auction block? If it is, then fillers can just strip cosigner once user's auction block is past

marktoda (approved):

src/reactors/PriorityOrderReactor.sol
```diff
@@ -67,41 +63,33 @@ contract PriorityOrderReactor is BaseReactor {
...
+        ) {
+            verify(order.cosigner, order.cosignerDigest(orderHash), order.cosignature);
+
+            auctionStartBlock = order.cosignerData.auctionTargetBlock;
+        }
+
         /// revert if the resolved auctionStartBlock is in the future
-        /// must be the minimum(auctionStartBlock, cosignerData.auctionTargetBlock)
-        if (order.auctionStartBlock > block.number) {
+        if (auctionStartBlock > block.number) {
```
marktoda, line 92: nit: above you do `block.number < auctionStartBlock` which is same thing but flipped... unifying would be nicer to read

src/reactors/PriorityOrderReactor.sol
zhongeric: yea but in that case we don't want to resolve the auctionStartBlock to the cosigned one, even if it's in the order and was signed over correctly

ConjunctiveNormalForm (approved):

src/reactors/PriorityOrderReactor.sol
zhongeric: oh I see what you're saying that fillers can just omit it from the order when submitting on chain and it's implicitly enforced by checking against block.number

PR #259, opened 2024-07-17
feat: revert quickly if order is already filled

zhongeric:
Thoughts on re-using the Permit2 custom error here of InvalidNonce? Gas savings are pretty decent, in the case where you would have reverted since the order is already filled, you save 10k gas


src/reactors/PriorityOrderReactor.sol
```diff
@@ -21,6 +21,8 @@ contract PriorityOrderReactor is BaseReactor {
     error InvalidDeadline();
     /// @notice thrown when an order's auctionStartBlock is in the future
     error OrderNotFillable();
+    /// @notice thrown when an order's nonce is invalid
+    error InvalidNonce();
```
marktoda, line 25: Maybe AlreadyFilled is better here? Though could also be canceled

marktoda (approved):
Maybe worth adding a few test cases for different nonce positions now to double check the bit manipulation, i.e. nonces used from diff words, other nonces in the same word etc

src/reactors/PriorityOrderReactor.sol
```diff
@@ -64,11 +66,14 @@ contract PriorityOrderReactor is BaseReactor {
     }
 
     /// @notice validate the priority order fields
+    /// - order must not be filled already
     /// - deadline must be in the future
     /// - resolved auctionStartBlock must not be in the future
     /// - if input scales with priority fee, outputs must not scale
     /// @dev Throws if the order is invalid
     function _validateOrder(bytes32 orderHash, PriorityOrder memory order) internal view {
```
marktoda, line 74: Seems we don't actually use orderHash here? So can hash the order after calling this to save bit more gas

marktoda: Oh nvm we do use it for cosigner digest

zhongeric: Ooo that saved 5k gas!

src/reactors/PriorityOrderReactor.sol
zhongeric: Yea I think AlreadyFilled is good

test/reactors/PriorityOrderReactor.t.sol
```diff
@@ -1,6 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0-or-later
 pragma solidity ^0.8.0;
 
+import {console2} from "forge-std/console2.sol";
```
marktoda, line 4: rm

marktoda (approved):

src/reactors/PriorityOrderReactor.sol
marktoda: sick!

PR #267, opened 2024-08-08
feat: Add priorityOrderReactor Quoter

zhongeric:


src/reactors/PriorityOrderReactorQuoter.sol
```diff
@@ -0,0 +1,142 @@
...
+                priorityFee = 0;
+            }
+        }
+    }
+
+    /// @notice check if an order has already been filled
+    /// @dev implementation copied from https://github.com/Uniswap/permit2/blob/cc56ad0f3439c502c246fc5cfcc3db92bb8b7219/src/SignatureTransfer.sol#L150
+    /// @param swapper the address of the swapper
+    /// @param nonce the nonce associated with the order
+    function _checkPermit2Nonce(address swapper, uint256 nonce) internal view {
```
ConjunctiveNormalForm, line 134: maybe worth moving this to a library, now that it's used in both the reactor and quoter?

src/reactors/PriorityOrderReactorQuoter.sol
```diff
@@ -0,0 +1,142 @@
...
+    error InputOutputScaling();
+    /// @notice thrown when tx gasprice is less than block.basefee
+    error InvalidGasPrice();
+
+    mapping(bytes32 => uint256) public orderAuctionStartBlocks;
+
+    constructor(IPermit2 _permit2, address _protocolFeeOwner) BaseReactor(_permit2, _protocolFeeOwner) {}
+
+    /// @inheritdoc BaseReactor
+    function _resolve(SignedOrder calldata signedOrder)
```
ConjunctiveNormalForm, line 38: probably need a public quote function? Can have a quoter interface that both this and the exiting quoter implements

PR #269, opened 2024-08-23
Non-linear Block-based Orders

codyborn:
- [x] New reactor, order type, and decay library to support non-linear block-based decay
- [x] Util lib
- [x] Util tests
- [x] NonLinearDecay Tests
- [x] Reactor tests

Some notable differences from Dutchv2 Orders:

1. Instead of an `uint256 endAmount`, input and output use a `NonLinearDecay curve` struct, which defines the relative points on the decay curve. The points are relative so that the cosigner only needs to override the `startAmount` and the user’s signed curve can remain unchanged.
2. We use `uint16` to store the relative block numbers. With 16 bits, we can define curves with points far into the future (11 days on Arbitrum and 1.5 years on Mainnet). If we were to use 8 bits, we could only define curves with 64 seconds into the future on Arbitrum.
3. `NonLinearDutchInput` now contains a `maxAmount` property to hold the max point on the input curve. Used for constructing the `InputToken` object for Permit2.


src/lib/ExclusivityLib.sol
```diff
@@ -54,7 +55,10 @@ library ExclusivityLib {
     /// @dev if the order has no exclusivity, always returns true
     /// @dev if the order has active exclusivity and the current filler is the exclusive address, returns true
     /// @dev if the order has active exclusivity and the current filler is not the exclusive address, returns false
-    function hasFillingRights(address exclusive, uint256 exclusivityEndTime) internal view returns (bool) {
-        return exclusive == address(0) || block.timestamp > exclusivityEndTime || exclusive == msg.sender;
+    function hasFillingRights(address exclusive, uint256 exclusivityEnd, bool timeBased) internal view returns (bool) {
+        return exclusive == address(0) ||
+            (timeBased && block.timestamp > exclusivityEnd) ||
+            (!timeBased && block.number > exclusivityEnd) ||
```
marktoda, line 61: if doing the above can just pass in `block.timestamp` or `block.number` as as `exclusivityCheck` param or something

src/lib/Util.sol
```diff
@@ -0,0 +1,45 @@
...
+            if (b < uint256(a)) {
+                revert NegativeUint();
+            }
+
+            return b - uint256(a);
+        }
+    }
+
+    // Retrieve the nth uint16 value from a packed uint256
+    function getUint16FromPacked(uint256 packedData, uint256 n) public pure returns (uint16) {
```
marktoda, line 23: Consider using a custom type a la https://github.com/Uniswap/v4-core/blob/main/src/types/BalanceDelta.sol

to represent 16xuint16s

src/lib/ExclusivityLib.sol
```diff
@@ -20,16 +20,17 @@ library ExclusivityLib {
...
+    /// @param exclusivityEnd The exclusivity end time/block
     /// @param exclusivityOverrideBps The exclusivity override BPS
     function handleExclusiveOverride(
         ResolvedOrder memory order,
         address exclusive,
-        uint256 exclusivityEndTime,
-        uint256 exclusivityOverrideBps
+        uint256 exclusivityEnd,
+        uint256 exclusivityOverrideBps,
+        bool timeBased
```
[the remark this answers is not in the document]
marktoda: also helps w/ readability of the params so we can make it clear what they're meant to be

src/lib/NonLinearDutchDecayLib.sol
```diff
@@ -0,0 +1,95 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+import {OutputToken, InputToken} from "../base/ReactorStructs.sol";
+import {NonLinearDutchOutput, NonLinearDutchInput, NonLinearDecay} from "../lib/NonLinearDutchOrderLib.sol";
+import {Util} from "./Util.sol";
+import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
+
+/// @notice helpers for handling dutch order objects
```
marktoda, line 9: nonlinear dutch order objects?

src/lib/Util.sol
```diff
@@ -0,0 +1,45 @@
+pragma solidity ^0.8.0;
+
+library Util {
+    error NegativeUint();
+    error IndexOutOfBounds();
+    error InvalidArrLength();
+
+    function subIntFromUint(int256 a, uint256 b) public pure returns (uint256) {
```
marktoda, line 8: i think you can do operator overloading so this is usable via a `-`, if not then `using for` at least

src/lib/NonLinearDutchDecayLib.sol
```diff
@@ -0,0 +1,95 @@
...
+    function decay(NonLinearDecay memory curve, uint256 startAmount, uint256 decayStartBlock)
+        internal
+        view
+        returns (uint256 decayedAmount)
+    {
+        // handle current block before decay or no decay
+        if (decayStartBlock >= block.number) {
+            return startAmount;
+        }
+        uint256 blockDelta = block.number - decayStartBlock;
```
marktoda, line 23: readability nit: consider separating out the process of finding the block/amount bounds from the process of calculating decay i.e. a helper `getBlockBounds(NonLinearDecay) returns (startBlock, endBlock, startAmount, endAmount)`?

src/lib/NonLinearDutchOrderLib.sol
```diff
@@ -0,0 +1,168 @@
...
+    bytes cosignature;
+}
+
+/// @dev The changes in tokens (positive or negative) to subtract from the start amount
+/// @dev The relativeBlocks should be strictly increasing
+struct NonLinearDecay {
+    // 16 uint16 values packed
+    // Can represent curves with points 2^16 blocks into the future
+    uint256 relativeBlocks;
+    int256[] relativeAmount;
```
marktoda, line 43: relativeAmounts?

marktoda (commented):
Looks great! Some thoughts

src/reactors/NonLinearDutchOrderReactor.sol
```diff
@@ -0,0 +1,141 @@
...
+
+/// @notice Reactor for non-linear dutch orders
+/// @dev V2 orders must be cosigned by the specified cosigner to override timings and starting values
+/// @dev resolution behavior:
+/// - If cosignature is invalid or not from specified cosigner, revert
+/// - If inputAmount is 0, then use baseInput
+/// - If inputAmount is nonzero, then ensure it is less than specified baseInput and replace startAmount
+/// - For each outputAmount:
+///   - If amount is 0, then use baseOutput
+///   - If amount is nonzero, then ensure it is greater than specified baseOutput and replace startAmount
```
marktoda, line 26: is this comment still just the v2 one? seems it could use some extra nonlinear info

src/lib/NonLinearDutchOrderLib.sol
```diff
@@ -0,0 +1,168 @@
...
+    // generic order information
+    OrderInfo info;
+    // The address which must cosign the full order
+    address cosigner;
+    // The tokens that the swapper will provide when settling the order
+    NonLinearDutchInput baseInput;
+    // The tokens that must be received to satisfy the order
+    NonLinearDutchOutput[] baseOutputs;
+    // signed over by the cosigner
+    CosignerData cosignerData;
```
marktoda, line 32: since these are purely relative, the cosigner now has full power to set the start amount to whatever they want? some thoughts here:
- this is a lot more power than cosigner has had in the past. Maybe we want to consider setting a min/max value?
- potentially cosigned start/end amounts could be so high/low to cause under/overflow after adjustment - probably want to make sure we cover that case

src/lib/NonLinearDutchDecayLib.sol
```diff
@@ -0,0 +1,95 @@
...
+        // iterate through the points and locate the current segment
+        for (uint256 i = 0; i < curve.relativeAmount.length; i++) {
```
zhongeric, line 25: also because the max number of points on the curve is known and strictly increasing, we can optimize this via a binary search once we know if `blockDelta` is greater than or equal to the mid value in the curve

src/reactors/NonLinearDutchOrderReactor.sol
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+    }
+
+    /// @notice validate the dutch order fields
+    /// - decay curves are defined
+    /// - deadline must have not passed
+    /// @dev Throws if the order is invalid
+    function _validateOrder(bytes32 orderHash, NonLinearDutchOrder memory order) internal pure {
+        if (order.baseInput.curve.relativeAmount.length == 0 || order.baseInput.curve.relativeAmount.length > 16) {
```
zhongeric, line 124: so this requires that you have a curve with `relativeAmount: [0]` for the case where you are applying no curve? Why not just accept the [] case

src/reactors/NonLinearDutchOrderReactor.sol
```diff
@@ -0,0 +1,141 @@
...
+            revert InvalidDecayCurve();
+        }
+        for (uint256 i = 0; i < order.baseOutputs.length; i++) {
+            if (order.baseOutputs[i].curve.relativeAmount.length == 0) {
```
zhongeric, line 128: also this check is inconsistent with the above one which checks the max length of relativeAmount

src/lib/Util.sol
codyborn: Defining a custom type for the test case `testGetUint16FromPacked` increases gas cost by 356. Do you think the readibility is worth the gas trade-off?

src/lib/Util.sol
codyborn: Unfortunately to overload the `-` both values need to be of the same type. I've added a new `sub` for uint256 types.

src/reactors/NonLinearDutchOrderReactor.sol
codyborn: Yup these are all still the same for non-linear orders. Just updated the header comment.

src/lib/NonLinearDutchOrderLib.sol
codyborn: The cosigner can only update the startAmount (not the relative amounts) and only if it's better than what the swapper signed ([see here](https://github.com/Uniswap/UniswapX/blob/295575ff548bc743e5a982e1a0339b268f350920/src/reactors/NonLinearDutchOrderReactor.sol#L95)). The curve and base amounts are set by the swapper. Because of this, I don't think the cosigner has more trust than it did in the past. LMK if I'm missing something.

If the cosigner is malicious and attempts to cause overflow, the tx will revert ([as it's currently implemented](https://github.com/Uniswap/UniswapX/blob/295575ff548bc743e5a982e1a0339b268f350920/test/lib/NonLinearDutchDecayLib.t.sol#L250)). Underflow isn't possible since we're subtracting an `int256` from a `uint256`. In the case of overflow we could either:
1. Do nothing and have the swapper deal with the failed tx (eg. use another cosigner).
2. Detect the overflow in the reactor and fallback to using the user provided input. This will be more gas intensive since we'd need to loop through the curve and check each point. 

Given this could only occur with a malicious cosigner and the worst case is that the tx reverts, I think it's safe to just let the tx revert.

src/reactors/NonLinearDutchOrderReactor.sol
codyborn: Good point. I was thinking there would always be some decay, but we can support limit orders too!

src/reactors/NonLinearDutchOrderReactor.sol
codyborn: Just moved these to `src/lib/NonLinearDutchDecayLib.sol` to save on gas (we're looping through outputs there already).

src/lib/NonLinearDutchDecayLib.sol
codyborn: I've implemented this in this [commit](https://github.com/Uniswap/UniswapX/pull/269/commits/59153044b910bb6d920f25cfc25d2469648421ad) however, I think it actually costs more gas for the length of points we're dealing with.
## Linear Search
image

## Binary Search
image

codyborn: Performed more extensive gas testing comparing linear with binary searching. Binary search only starts showing real benefits when we have 14 or more curve points. Opting for linear search since we will likely have less than 5 curve points defined and it's less bug prone.

image

marktoda:
seems you need a `forge fmt`

marktoda (commented):
any tests for the reactor as a whole?

src/lib/V3DutchDecayLib.sol
```diff
@@ -0,0 +1,145 @@
...
+import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
+import {sub} from "./MathExt.sol";
+import {Uint16Array, fromUnderlying} from "../types/Uint16Array.sol";
+
+/// @notice thrown when the decay curve is invalid
+error InvalidDecayCurve();
+
+/// @notice helpers for handling non-linear dutch order objects
+library V3DutchDecayLib {
+
```
marktoda, line 15: nit: rm \n

src/lib/V3DutchDecayLib.sol
```diff
@@ -0,0 +1,145 @@
...
+import {V3DutchOutput, V3DutchInput, V3Decay} from "../lib/V3DutchOrderLib.sol";
+import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
+import {sub} from "./MathExt.sol";
+import {Uint16Array, fromUnderlying} from "../types/Uint16Array.sol";
+
+/// @notice thrown when the decay curve is invalid
+error InvalidDecayCurve();
+
+/// @notice helpers for handling non-linear dutch order objects
+library V3DutchDecayLib {
```
marktoda, line 14: consider naming NonlinearDutchDecayLib for specificity here - generally I see these libraries as independent from the reactor implementations they service, where reactors are just a specific aggregation of features from libraries

src/lib/V3DutchDecayLib.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// @return next The relative block after the current position
+    function locateCurvePosition(V3Decay memory curve, uint16 currentRelativeBlock)
+        internal
+        pure
+        returns (uint16 prev, uint16 next)
+    {
+        Uint16Array relativeBlocks = fromUnderlying(curve.relativeBlocks);
+        prev = 0;
+        next = 0;
+        while(next < curve.relativeAmounts.length) {
```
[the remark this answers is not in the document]
marktoda: nit: while -> for? more clear if/when it terminates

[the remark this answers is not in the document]
marktoda: also I think unchecked increments only applies to the incrementor in a loop

src/lib/V3DutchDecayLib.sol
```diff
@@ -0,0 +1,145 @@
...
+        return (next - 1, next - 1);
+    }
+
+    /// @notice returns the linear interpolation between the two points
+    /// @param startBlock The start of the decay
+    /// @param endBlock The end of the decay
+    /// @param currentBlock The current position in the decay
+    /// @param startAmount The amount of the start of the decay
+    /// @param endAmount The amount of the end of the decay
+    function linearDecay(
```
marktoda, line 81: could this be refactored to share logic with the linear decay in `DutchDecayLib`?

src/reactors/V3DutchOrderReactor.sol
```diff
@@ -0,0 +1,127 @@
...
+        view
+        virtual
+        override
+        returns (ResolvedOrder memory resolvedOrder)
+    {
+        V3DutchOrder memory order = abi.decode(signedOrder.order, (V3DutchOrder));
+        // hash the order _before_ overriding amounts, as this is the hash the user would have signed
+        bytes32 orderHash = order.hash();
+
+        if (order.info.deadline < block.timestamp) {
```
marktoda, line 60: can move this into validation fn?

src/types/Uint16Array.sol
```diff
@@ -0,0 +1,44 @@
...
+function toUint256(uint16[] memory inputArray) pure returns (uint256 uint16Array) {
+    return Uint16Array.unwrap(toUint16Array(inputArray));
+}
+
+function fromUnderlying(uint256 value) pure returns (Uint16Array) {
+    return Uint16Array.wrap(value);
+}
+
+// Helper for creating a packed uint256 from a uint16 array
+function toUint16Array(uint16[] memory inputArray) pure returns (Uint16Array uint16Array) {
```
marktoda, line 19: imo better to keep these in the library, global functions are a bit awkward and not standard for uniswap style. Generally the `using for` functions at the top of the library and any extraneous helpers at the bottom

src/lib/V3DutchDecayLib.sol
```diff
@@ -0,0 +1,145 @@
...
+        Uint16Array relativeBlocks = fromUnderlying(curve.relativeBlocks);
+        uint16 blockDelta = uint16(block.number - decayStartBlock);
+        // Special case for when we need to use the decayStartBlock (0)
+        if (relativeBlocks.getElement(0) > blockDelta) {
+            return linearDecay(0, relativeBlocks.getElement(0), blockDelta, startAmount, startAmount.sub(curve.relativeAmounts[0]));
+        }
+        // the current pos is within or after the curve
+        uint16 prev;
+        uint16 next;
+        (prev, next) = locateCurvePosition(curve, blockDelta);
```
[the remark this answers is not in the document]
codyborn: Sweet!

src/lib/V3DutchDecayLib.sol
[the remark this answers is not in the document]
codyborn: Sweet! I was able to drop the most comprehensive test by an avg of `534` gas with these tips.

src/reactors/V3DutchOrderReactor.sol
codyborn: Yeah, I just need to make it a `view` but it works.

test/reactors/V3DutchOrderReactor.t.sol
```diff
@@ -0,0 +1,1005 @@
...
+        uint256 overriddenInputAmount = 0.7 ether;
+        tokenIn.mint(swapper, overriddenInputAmount);
+        tokenOut.mint(address(fillContract), outputAmount);
+        tokenIn.forceApprove(swapper, address(permit2), type(uint256).max);
+        CosignerData memory cosignerData = CosignerData({
+            decayStartBlock: block.number,
+            exclusiveFiller: address(0),
+            exclusivityOverrideBps: 0,
+            inputAmount: overriddenInputAmount,
+            outputAmounts: ArrayBuilder.fill(1, 0)
```
zhongeric, line 196: nit: not sure if this is the same from V2DutchOrder tests but the gas on this will be off due to the 0 transfer of the output

test/reactors/V3DutchOrderReactor.t.sol
```diff
@@ -0,0 +1,1005 @@
...
+        _snapStart("V3-ExclusiveFiller");
+        fillContract.execute(signedOrder);
+        snapEnd();
+
+        assertEq(tokenIn.balanceOf(swapper), 0);
+        assertEq(tokenOut.balanceOf(swapper), outputAmount);
+        assertEq(tokenIn.balanceOf(address(fillContract)), inputAmount);
+    }
+
+    function testV3ExclusiveOverrideInvalidCallerCosignedAmountOutput() public {
```
zhongeric, line 327: this is not invalidCaller since it succeeds? Maybe like "testV3AppliesExclusiveOverride..."

test/reactors/V3DutchOrderReactor.t.sol
```diff
@@ -0,0 +1,1005 @@
...
+                    outputAmount,
+                    CurveBuilder.singlePointCurve(1000, int256(inputAmount * 10 / 100)),
+                    address(swapper)
+                )
+            })
+        );
+        uint256 swapperInputBalanceStart = tokenIn.balanceOf(address(swapper));
+        uint256 swapperOutputBalanceStart = tokenOut.balanceOf(address(swapper));
+        vm.expectEmit(false, true, true, false, address(reactor));
+        emit Fill(keccak256("not checked"), address(fillContract), swapper, 0);
```
zhongeric, line 552: can we not check against the local order.hash() here?

zhongeric (changes requested):

src/lib/NonLinearDutchDecayLib.sol
```diff
@@ -7,40 +7,88 @@ import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
...
+        // Special case for when we need to use the decayStartBlock (0)
+        if (relativeBlocks.getElement(0) > blockDelta) {
+            return linearDecay(0, relativeBlocks.getElement(0), blockDelta, startAmount, startAmount.sub(curve.relativeAmounts[0]));
+        }
+        // the current pos is within or after the curve
+        uint16 prev;
+        uint16 next;
+        (prev, next) = locateCurvePosition(curve, blockDelta);
+        uint256 lastAmount = startAmount.sub(curve.relativeAmounts[prev]);
+        uint256 nextAmount = startAmount.sub(curve.relativeAmounts[next]);
```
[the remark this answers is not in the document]
codyborn: Yup, the curve points are type `int256` so we can move it in either direction (including negative decay).

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -0,0 +1,119 @@
...
+import {FixedPointMathLib} from "solmate/src/utils/FixedPointMathLib.sol";
+import {sub} from "./MathExt.sol";
+import {Uint16ArrayLibrary, Uint16Array, fromUnderlying} from "../types/Uint16Array.sol";
+import {DutchDecayLib} from "./DutchDecayLib.sol";
+
+/// @notice thrown when the decay curve is invalid
+error InvalidDecayCurve();
+
+/// @notice helpers for handling non-linear dutch order objects
+library NonlinearDutchDecayLib {
```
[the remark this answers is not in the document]
codyborn: Yeah, Mark suggested "Nonlinear" over "V3" for the decay lib.

src/lib/V3DutchOrderLib.sol
```diff
@@ -94,7 +94,7 @@ library NonLinearDutchOrderLib {
     /// @dev Note that sub-structs have to be defined in alphabetical order in the EIP-712 spec
     string internal constant PERMIT2_ORDER_TYPE = string(
         abi.encodePacked(
-            "NonLinearDutchOrder witness)",
+            "V3DutchOrder witness)",
```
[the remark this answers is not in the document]
codyborn: Ah thank you.. I did a find/replace and forgot to update the ordering!

src/lib/V3DutchOrderLib.sol
```diff
@@ -103,18 +103,18 @@ library NonLinearDutchOrderLib {
...
     }
 
     /// @notice hash the given input
     /// @param input the input to hash
     /// @return the eip-712 input hash
-    function hash(NonLinearDutchInput memory input) internal pure returns (bytes32) {
+    function hash(V3DutchInput memory input) internal pure returns (bytes32) {
         return keccak256(
             abi.encode(
                 NON_LINEAR_DUTCH_OUTPUT_TYPE_HASH, input.token, input.startAmount, hash(input.curve), input.maxAmount
```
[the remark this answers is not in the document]
codyborn: Yes, good catch!

src/reactors/V3DutchOrderReactor.sol
```diff
@@ -117,19 +104,13 @@ contract NonLinearDutchOrderReactor is BaseReactor {
...
-                revert InvalidDecayCurve();
-            }
+    function _validateOrder(bytes32 orderHash, V3DutchOrder memory order) internal view {
+        if (order.info.deadline < block.timestamp) {
+            revert DeadlineReached();
         }
-
         (bytes32 r, bytes32 s) = abi.decode(order.cosignature, (bytes32, bytes32));
         uint8 v = uint8(order.cosignature[64]);
         // cosigner signs over (orderHash || cosignerData)
```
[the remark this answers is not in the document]
codyborn: Sweet, good idea!

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -0,0 +1,119 @@
...
+    using FixedPointMathLib for uint256;
+    using {sub} for uint256;
+    using Uint16ArrayLibrary for Uint16Array;
+
+    /// @notice locates the current position on the curve and calculates the decay
+    /// @param curve The curve to search
+    /// @param startAmount The absolute start amount
+    /// @param decayStartBlock The absolute start block of the decay
+    function decay(NonlinearDutchDecay memory curve, uint256 startAmount, uint256 decayStartBlock)
```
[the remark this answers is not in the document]
codyborn: Yeah I was worried about the additional gas for verifying in the contract and instead we can verify in the SDK.

test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -2,133 +2,193 @@
...
+
+        snapStart("V3-LocateCurvePositionSingle");
+        (uint16 prev, uint16 next) = NonlinearDutchDecayLib.locateCurvePosition(curve, 1);
+        assertEq(prev, 0);
+        assertEq(next, 0);
+
+        (prev, next) = NonlinearDutchDecayLib.locateCurvePosition(curve, 2);
+        assertEq(prev, 0);
+        assertEq(next, 0);
+        snapEnd();
```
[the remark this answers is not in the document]
codyborn: Yeah, since each call can have different gas, I've been measuring the per-call gas by taking the avg over a series of calls (e.g. `.forge-snapshots/V3-ExtendedMultiPointDutchDecay.snap` / 21).

test/reactors/V3DutchOrderReactor.t.sol
codyborn: [Yeah it's similar to V2](https://github.com/Uniswap/UniswapX/blob/b3a807b589a2b8fe2220fdab44ea66eaf2ce2e7e/test/reactors/V2DutchOrderReactor.t.sol#L364). I've been using the gas measurements more just to get a sense of the relative changes.

test/reactors/V3DutchOrderReactor.t.sol
codyborn: I believe this is saying that the `vm.expectEmit()` doesn't check the value here (I took this from the V2 tests).

src/lib/MathExt.sol
```diff
@@ -0,0 +1,85 @@
...
+    /// @custom:throws NegativeUint if the subtraction would result in a negative value.
+    function sub(uint256 a, int256 b) internal pure returns (uint256) {
+        if (b < 0) {
+            // If b is negative, add its absolute value to a
+            return a + uint256(-b);
+        } else {
+            // If b is positive, subtract it from a
+            if (a < uint256(b)) {
+                revert NegativeUint();
+            }
```
marktoda, line 22: nit: this should be implicit in checked subtraction, no?

src/lib/MathExt.sol
```diff
@@ -0,0 +1,85 @@
...
+    /// @param max The maximum bound for the result.
+    /// @return r The result of the bounded subtraction.
+    function boundedSub(uint256 a, int256 b, uint256 min, uint256 max) internal pure returns (uint256 r) {
+        if (b < 0) {
+            // If b is negative, add its absolute value to a
+            uint256 absB = uint256(-b);
+            // would overflow
+            if (type(uint256).max - absB < a) {
+                return max;
+            }
```
marktoda, line 51: should be captured in checked math?

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -0,0 +1,126 @@
...
+        }
+
+        // handle current block before decay or no decay
+        if (decayStartBlock >= block.number || curve.relativeAmounts.length == 0) {
+            return startAmount.bound(minAmount, maxAmount);
+        }
+        Uint16Array relativeBlocks = fromUnderlying(curve.relativeBlocks);
+        uint16 blockDelta = uint16(block.number - decayStartBlock);
+        int256 curveDelta;
+        // Special case for when we need to use the decayStartBlock (0)
```
marktoda, line 44: I think this comment could use a little more color, taking me some effort to figure out what's happening here

So decayStartBlock is already past, but the first block delta hasn't been hit yet? So you're anchoring the left end on block 0? I'm not sure I see the reason even for linear decay there, since it's just going to be sooo close to startAmount + relativeAmounts[0] right?

marktoda: edit: ohh it's _blockDelta_ of 0, not block 0 right?

marktoda: this is one case where maybe named parameters for linearDecay would help

marktoda: Possible to just contain this case within locateCurvePosition?

marktoda (dismissed):
lgtm! Few ideas

src/lib/MathExt.sol
codyborn: Yes, but I do want to return the `max` value instead of reverting.

src/lib/NonlinearDutchDecayLib.sol
codyborn: Just refactored it.. lmk what you thnk!

marktoda (approved):

PR #273, opened 2024-09-11
Gas Adjustment Feature

codyborn:
Adjusts the auction start price based on the change in `block.basefee` from the time the swapper signed the order to the time of execution. The adjustment can be in favor of the swapper (unlimited) or against the swapper (bounded by slippage).

The goal of this mainnet-focused feature is to:
- improve PI (give users more tokens when gas decreases)
- decrease fade rate and missing-exclusivity when gas prices increase

To know how much to adjust the input/output token by, I introduce a new field `adjustmentPerGweiBasefee` which is the amount of token to change per gwei of change to the basefee. As @0age mentioned on the call, it should be calculated by (total gas of swap * token/eth).


src/lib/MathExt.sol
```diff
@@ -1,17 +1,52 @@
...
+            // If b is positive, subtract it from a
+            if (a < uint256(b)) {
+                revert NegativeUint();
+            }
+
+            return a - uint256(b);
+        }
+    }
+
+    function boundedSub(uint256 a, int256 b, uint256 min, uint256 max) internal pure returns (uint256 r) {
```
zhongeric, line 23: i think it would be nice to introduce another function in this library "bound", which applies the bounding logic at the end of this function

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -14,42 +14,49 @@ error InvalidDecayCurve();
...
+    ) internal view returns (uint256 decayedAmount) {
         // mismatch of relativeAmounts and relativeBlocks
         if (curve.relativeAmounts.length > 16) {
             revert InvalidDecayCurve();
         }
 
         // handle current block before decay or no decay
         if (decayStartBlock >= block.number || curve.relativeAmounts.length == 0) {
-            return startAmount;
+            return startAmount.boundedSub(0, minAmount, maxAmount);
```
zhongeric, line 38: think its cleaner to do startAmount.bounded(minAmount, maxAmount) here

src/lib/V3DutchOrderLib.sol
```diff
@@ -28,6 +28,8 @@ struct V3DutchOrder {
     V3DutchInput baseInput;
     // The tokens that must be received to satisfy the order
     V3DutchOutput[] baseOutputs;
+    // Adjust the startAmount to account for changes to gas
+    uint256 gasSnapshot;
```
zhongeric, line 32: something more descriptive that specifies what unit this is in would be nice, like `baseFeeGwei `

src/lib/V3DutchOrderLib.sol
```diff
@@ -65,41 +69,60 @@ struct V3DutchOutput {
...
     using OrderInfoLib for OrderInfo;
 
-    bytes internal constant NON_LINEAR_DUTCH_ORDER_TYPE = abi.encodePacked(
+    bytes internal constant V3_DUTCH_ORDER_TYPE = abi.encodePacked(
         "V3DutchOrder(",
         "OrderInfo info,",
         "address cosigner,",
         "V3DutchInput baseInput,",
-        "V3DutchOutput[] baseOutputs)"
+        "V3DutchOutput[] baseOutputs," "uint256 gasSnapshot)"
```
zhongeric, line 87: missing comma here and fmt

src/lib/V3DutchOrderLib.sol
```diff
@@ -53,6 +55,8 @@ struct V3DutchInput {
     NonlinearDutchDecay curve;
     // The max amount of the curve
     uint256 maxAmount;
+    // The amount of token to change per wei change in basefee
+    uint256 adjustmentPerGweiBasefee;
```
zhongeric, line 59: BaseFee?

src/lib/V3DutchOrderLib.sol
zhongeric: ah also I think this should be moved above the baseInput and baseOutputs, it's easier for parsing to have all of the statically sized elements before any dynamic ones IMO

src/reactors/V3DutchOrderReactor.sol
```diff
@@ -103,6 +109,23 @@ contract V3DutchOrderReactor is BaseReactor {
         }
     }
 
+    function _updateWithGasAdjustment(V3DutchOrder memory order) internal view {
+        // positive means an increase in gas
+        int256 gasDeltaGwei = block.basefee.sub(order.gasSnapshot);
+
+        // Gas increase should increase input
+        int256 inputDelta = int256(order.baseInput.adjustmentPerGweiBasefee) * gasDeltaGwei / 1 gwei;
```
zhongeric, line 117: is gwei too large of a denominator? I wonder if wei or some percentage of a gwei would be better, then we can move this into a constant

src/reactors/V3DutchOrderReactor.sol
```diff
@@ -103,6 +109,23 @@ contract V3DutchOrderReactor is BaseReactor {
...
+        order.baseInput.startAmount =
+            order.baseInput.startAmount.boundedSub(0 - inputDelta, 0, order.baseInput.maxAmount);
```
zhongeric, line 119: I would prefer doing boundedAdd vs. boundedSub, or doing add / sub and calling `bounded(min, max)` afterwards. Passing a negative value into a subtraction function feels like an antipattern

src/lib/DutchDecayLib.sol
```diff
@@ -68,6 +68,35 @@ library DutchDecayLib {
         }
     }
 
+    /// @notice returns the linear interpolation between the two points
+    /// @param startPoint The start of the decay
+    /// @param endPoint The end of the decay
+    /// @param currentPoint The current position in the decay
+    /// @param startAmount The amount of the start of the decay
+    /// @param endAmount The amount of the end of the decay
+    function linearDecay(
```
codyborn, line 77: I added another version of this that accepts int256 for start and end amount.
1. I wanted to decay only through the relative amounts so that I can apply the bounding logic (min/max) check after the decay.
2. It was too convoluted to try to support both in the same function. 

Note that this linearDecay performs `mulDivDown` in both positive and negative cases as it felt more intuitive. Happy to discuss!

src/lib/V3DutchOrderLib.sol
```diff
@@ -65,41 +69,60 @@ struct V3DutchOutput {
...
     string internal constant PERMIT2_ORDER_TYPE = string(
         abi.encodePacked(
             "V3DutchOrder witness)",
             NON_LINEAR_DECAY_TYPE,
-            NON_LINEAR_DUTCH_ORDER_TYPE,
-            NON_LINEAR_DUTCH_OUTPUT_TYPE,
             OrderInfoLib.ORDER_INFO_TYPE,
-            DutchOrderLib.TOKEN_PERMISSIONS_TYPE
+            DutchOrderLib.TOKEN_PERMISSIONS_TYPE,
+            V3_DUTCH_INPUT_TYPE,
```
codyborn, line 123: Some of these are fixes from this PR
https://github.com/Uniswap/UniswapX/pull/269

src/lib/DutchDecayLib.sol
```diff
@@ -68,6 +68,35 @@ library DutchDecayLib {
...
+        uint256 endPoint,
+        uint256 currentPoint,
+        int256 startAmount,
+        int256 endAmount
+    ) internal pure returns (int256) {
+        if (currentPoint >= endPoint) {
+            return endAmount;
+        }
+        unchecked {
+            uint256 elapsed = currentPoint - startPoint;
```
marktoda, line 88: do we need to check for startPoint > currentPoint to avoid overflow here?

src/lib/DutchDecayLib.sol
marktoda: can the uint one call this one w/ casted params?

src/reactors/V3DutchOrderReactor.sol
marktoda: mulDivDown? or does that not support signed ints

src/lib/V3DutchOrderLib.sol
```diff
@@ -128,11 +156,13 @@ library V3DutchOrderLib {
...
             abi.encode(
-                NON_LINEAR_DUTCH_OUTPUT_TYPE_HASH,
+                V3_DUTCH_OUTPUT_TYPE_HASH,
                 output.token,
                 output.startAmount,
                 hash(output.curve),
-                output.recipient
+                output.recipient,
+                output.minAmount,
+                output.adjustmentPerGweiBasefee
```
marktoda, line 165: is `gwei` enough granularity? think even partial gwei changes in basefee may be impactful

src/reactors/V3DutchOrderReactor.sol
codyborn: @marktoda yeah mulDivDown only supports uint256.

@zhongeric I was thinking about this a lot too.. I think gwei is the easiest to reason about with today's gas prices. The int256 gives us ~1e76 of space and if we assume gas is around 1e9-1e11 we get a range of 1 wei - 1e65 for possible `adjustmentPerGweiBasefee` values.

If we look at low value token like SHIB for example which has 18 decimals and is worth 0.0000000057064226 ETH, 1 gwei = 5.7064226 SHIB, so we'd want the `adjustmentPerGweiBasefee = 5706422600 * 10k gas` (ignoring tx call gas) and we'd still have 1e53 of wiggle room before overflowing.

src/lib/DutchDecayLib.sol
codyborn: Yup! I was trying to do the other way around, but as long as we're okay with losing 1 bit of precision, it's cleaner.

src/lib/V3DutchOrderLib.sol
codyborn: Thanks! I'll call it `baseFee` since it's the actual basefee denominated in wei.

src/lib/DutchDecayLib.sol
codyborn: Good question. It's not possible with the current code since the caller passes in either 0 or `relativeBlocks.getElement(prev)` for the `startPoint` where `relativeBlocks.getElement(prev)` is guaranteed to be less than the `currentPoint`.
Similarly it's not possible for the `startPoint` to be greater than or equal to the `endPoint`. It could be equal if (currentPoint > endPoint) but we check that above. Otherwise, the `locateCurvePosition()` call ensures that it halts the search as soon as it finds a block that is greater than or equal to the `blockDelta`, meaning every value that came before it would be less than the `blockDelta` (and transitively less than the block it finds).

src/lib/V3DutchOrderLib.sol
codyborn: Yeah, it's denominated in gwei but since the division by 1 gwei happens after the multiplication, we can take advantage of wei-level changes in the gas fee.
`int256 inputDelta = int256(order.baseInput.adjustmentPerGweiBaseFee) * gasDeltaGwei / 1 gwei`

See this test for example:
https://github.com/Uniswap/UniswapX/blob/c0d699883b02668aed1d95077efb4a78ec7d4cfc/test/reactors/V3DutchOrderReactor.t.sol#L1127

src/lib/DutchDecayLib.sol
```diff
@@ -17,6 +17,9 @@ library DutchDecayLib {
     /// @notice thrown if the endTime of an order is before startTime
     error EndTimeBeforeStartTime();
 
+    /// @notice thrown if the decay duration is zero
+    error ZeroDuration();
```
zhongeric, line 21: is this used?

src/lib/MathExt.sol
```diff
@@ -38,15 +59,27 @@ library MathExt {
...
+    /// @return The result of the subtraction as a signed integer.
     function sub(uint256 a, uint256 b) internal pure returns (int256) {
         if (a < b) {
             return 0 - SafeCast.toInt256(b - a);
         } else {
             return SafeCast.toInt256(a - b);
         }
     }
+
+    /// @notice Bounds a value between a minimum and maximum value.
```
zhongeric, line 77: nit: bounds a uint between a minimum and maximum value

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -8,19 +8,20 @@ import {MathExt} from "./MathExt.sol";
...
-error InvalidDecayCurve();
-
 /// @notice helpers for handling non-linear dutch order objects
 library NonlinearDutchDecayLib {
     using FixedPointMathLib for uint256;
     using MathExt for uint256;
     using Uint16ArrayLibrary for Uint16Array;
 
+    /// @notice thrown when the decay curve is invalid
+    error InvalidDecayCurve();
```
zhongeric, line 18: is this used?

src/lib/V3DutchOrderLib.sol
```diff
@@ -24,12 +24,12 @@ struct V3DutchOrder {
     OrderInfo info;
     // The address which must cosign the full order
     address cosigner;
+    // Adjust the startAmount to account for changes to gas
+    uint256 baseFee;
```
zhongeric, line 28: nit: something like `startingBaseFee` or `originalBaseFee` could be more descriptive

test/lib/DutchDecayLib.t.sol
```diff
@@ -74,27 +74,26 @@ contract DutchDecayLibTest is Test {
         assertEq(DutchDecayLib.decay(2 ether, 1 ether, 100, 200), 1 ether);
     }
 
-    function testDutchDecayBounded(uint256 startAmount, uint256 endAmount, uint256 decayStartTime, uint256 decayEndTime)
+    function testDutchDecayBounded(int256 startAmount, int256 endAmount, uint256 decayStartTime, uint256 decayEndTime)
```
zhongeric, line 77: why change types to int256 to cast to uint256 in the test?

zhongeric (approved):
looks good with some comments!

src/lib/NonlinearDutchDecayLib.sol
codyborn: Yes, line 34 of this file (if relativeAmounts length is outside of expected range).

test/lib/DutchDecayLib.t.sol
codyborn: it helps bound the values to a range to prevent overflow.. I tried `vm.assume(startAmount < 2* 255 - 1)` but forge testing would fail saying it hit the max number of failing input.

zhongeric: Ah I see, and bound would reject too many probably

zhongeric (approved):
Niceee!!

src/lib/DutchDecayLib.sol
codyborn: It IS possible for this int256 cast to overflow if `startAmount` is max int256 and `endAmount` is negative:
`int256(0 - uint256(startAmount - endAmount).mulDivDown(elapsed, duration)`

Going to remove the `unchecked`.

marktoda (approved):

PR #277, opened 2024-09-25
fix: invalidCosignature already defined in CosignerLib

alanhwu:
`invalidCosignature()` is never used as a revert in `V3DutchOrderReactor.sol`. It is already declared in `CosignerLib` where the revert based on cosignature occurs.


codyborn (approved):
Thanks!

PR #280, opened 2024-09-30
audit(14): arbitrum block number

zhongeric:


test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -10,21 +10,23 @@ import {V3DutchOutput, V3DutchInput, NonlinearDutchDecay} from "../../src/lib/V3
...
         uint256 minAmount,
         uint256 maxAmount
     ) public view {
-        NonlinearDutchDecayLib.decay(curve, startAmount, decayStartBlock, minAmount, maxAmount);
+        NonlinearDutchDecayLib.decay(curve, startAmount, decayStartBlock, blockNumberish, minAmount, maxAmount);
     }
 }
 
-contract NonlinearDutchDecayLibTest is Test, GasSnapshot {
+contract NonlinearDutchDecayLibTest is Test, GasSnapshot, BlockNumberish {
```
alanhwu, line 29: Wondering if we should do another test suite for chainId 42161 that uses `setBlockNumber` since our current tests just use `vm.roll()` which wouldn't do anything in an Arbitrum environment

alanhwu (approved):
Just a question about testing Arbitrum separately

test/lib/NonLinearDutchDecayLib.t.sol
zhongeric: Ahh yeah so imo the NonLinearDutchDecayLib should function completely agnostic of the behavior of BlockNumberish - since it strictly takes it in as a parameter. So a test that the blockNumberish parameter passed into the Lib is correctly set may be better in the reactor test

src/interfaces/IArbSys.sol
```diff
@@ -0,0 +1,13 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+/**
+ * @notice Minimal interface for interacting with Arbitrum system contracts
+ */
+interface IArbSys {
```
codyborn, line 7: We can add this as a submodule:
```
[submodule "lib/nitro-contracts"]
	path = lib/nitro-contracts
	url = https://github.com/OffchainLabs/nitro-contracts
```

src/base/BlockNumberish.sol
```diff
@@ -0,0 +1,31 @@
...
+        if (block.chainid == 42161) {
+            _getBlockNumberish = _getBlockNumberSyscall;
+        } else {
+            _getBlockNumberish = _getBlockNumber;
+        }
+    }
+
+    /// @dev Private function to get the block number on arbitrum
+    function _getBlockNumberSyscall() private view returns (uint256) {
+        return IArbSys(0x0000000000000000000000000000000000000064).arbBlockNumber();
```
codyborn, line 24: this address and `42161` can be constants at the top of this file:
```
address constant arbSysAddress = 0x0000000000000000000000000000000000000064;
uint256 constant arbChainId = 42161;
```

codyborn (approved):
Looks good! Left some small suggestions

src/interfaces/IArbSys.sol
zhongeric: Agree but not sure if necessary since we also don't need the entire IArbSys interface, just this one func

codyborn (dismissed):

codyborn (dismissed):

codyborn (approved):

PR #281, opened 2024-09-30
fix: EIP-712 nested types hash

alanhwu:
The V3 Dutch Input & Output Types have a nested custom type `NonlinearDutchDecay`, which must be concatenated when determining the Input/Output type hashes


marktoda (commented):
now that you know the expected typehashes from typescript, can add a small unit test to assert them

marktoda (approved):

PR #299, opened 2024-10-01
audit(1): Missing min < max check

alanhwu:
Require that `boundedSub` follows the invariant of: `min <= max`

### DIFF
--- a/src/lib/MathExt.sol
+++ b/src/lib/MathExt.sol
@@ -34,6 +34,7 @@ library MathExt {
     /// @param max The maximum bound for the result.
     /// @return r The result of the bounded subtraction.
     function boundedSub(uint256 a, int256 b, uint256 min, uint256 max) internal pure returns (uint256 r) {
+        require(min <= max, "MathExt: min must be less than or equal to max");
         if (b < 0) {
             // If b is negative, add its absolute value to a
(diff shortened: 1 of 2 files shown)


zhongeric (dismissed):
nice! two comments

zhongeric (dismissed):

zhongeric (approved):

src/lib/MathExt.sol
```diff
@@ -34,6 +34,7 @@ library MathExt {
     /// @param max The maximum bound for the result.
     /// @return r The result of the bounded subtraction.
     function boundedSub(uint256 a, int256 b, uint256 min, uint256 max) internal pure returns (uint256 r) {
+        require(min <= max, "MathExt: min must be less than or equal to max");
```
codyborn, line 37: I wonder if this is better to check in the SDK to save the additional gas.

codyborn: If we're going to keep it, we should switch to using a custom revert message:
https://www.notion.so/uniswaplabs/Protocols-Research-Team-Homepage-e55da9af622047b5b289b7b63e7d899e?pvs=4#b2048399e4664f6ebb1b6ddbec507a39

PR #304, opened 2024-10-02
audit(12): Long-term orders may be filled with an incorrect amount

alanhwu:
When the `blockDelta` is larger than uint16, we can face an overflow that will cause orders that decay past this point to have incorrect decayed amounts.


test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -474,4 +474,36 @@ contract NonlinearDutchDecayLibTest is Test, GasSnapshot {
         vm.expectRevert(NonlinearDutchDecayLib.InvalidDecayCurve.selector);
         mockNonlinearDutchDecayLibContract.decay(curve, startAmount, decayStartBlock, 1 ether, 1 ether);
     }
+
+    function testFuzzDutchDecayBeyondUint16Max(
+        uint256 startAmount,
+        uint256 blockDelta,
+        uint16 lastValidBlock,
+        uint256 decayAmountFuzz
```
zhongeric, line 482: Consider making the fuzzed inputs match the inputs to decay(), and using vm.assume to trigger the overflow condition

test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -474,4 +474,36 @@ contract NonlinearDutchDecayLibTest is Test, GasSnapshot {
...
+        uint256 decayStartBlock = 100;
+        vm.assume(lastValidBlock > 0);
+        vm.assume(startAmount > 0);
+
+        // Testing that we get a fully decayed curve instead of overflowed mistake
+        vm.assume(blockDelta > uint256(type(uint16).max));
+        vm.assume(blockDelta < type(uint256).max - decayStartBlock);
+
+        // Bound decayAmountFuzz between 0 and startAmount
+        decayAmountFuzz = bound(decayAmountFuzz, 0, startAmount);
```
zhongeric, line 493: nit: extract 0 into a named variable

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -39,7 +40,7 @@ library NonlinearDutchDecayLib {
             return startAmount.bound(minAmount, maxAmount);
         }
 
-        uint16 blockDelta = uint16(block.number - decayStartBlock);
```
zhongeric, line 42: add comment here about rationale for the change, and the expected behavior in case of overflow

test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -500,10 +507,12 @@ contract NonlinearDutchDecayLibTest is Test, GasSnapshot {
...
 
-        uint256 currentBlock = decayStartBlock + blockDelta;
         vm.roll(currentBlock);
-        uint256 decayed = NonlinearDutchDecayLib.decay(curve, startAmount, decayStartBlock, 0, startAmount);
-
-        assertEq(decayed, startAmount - decayAmountFuzz, "Should be fully decayed for block delta beyond uint16.max");
+        uint256 decayed = NonlinearDutchDecayLib.decay(curve, startAmount, decayStartBlock, minAmount, maxAmount);
+        assertEq(
+            decayed,
+            Math.max(startAmount - decayAmountFuzz, minAmount),
```
zhongeric, line 514: if we are restricting to the case where an overflow happens, shouldn't decayed always equal minAmount and nothing else?

alanhwu: You can set a `minAmount` that's lower than the last decay of a curve. In the extreme case someone could even put 0

zhongeric: ah yes because you also specify the relativeAmounts

zhongeric (dismissed):

codyborn (dismissed):

zhongeric (approved):

PR #305, opened 2024-10-02
audit(16): Round decays in favor of swapper

alanhwu:
To favor the swapper, decay outputs should be rounded up and decay inputs should be rounded down.


src/lib/DutchDecayLib.sol
```diff
@@ -84,6 +102,47 @@ library DutchDecayLib {
...
+                delta = int256(uint256(endAmount - startAmount).mulDivUp(elapsed, duration));
+            }
+        } else {
+            // For outputs, we want to minimize startAmount + delta to favor the swapper
+            if (endAmount < startAmount) {
+                delta = -int256(uint256(startAmount - endAmount).mulDivUp(elapsed, duration));
+            } else {
+                delta = int256(uint256(endAmount - startAmount).mulDivDown(elapsed, duration));
+            }
+        }
```
zhongeric, line 141: A suggestion I have for this is to condense the functions in this file into two distinct functions:
`linearDecayRoundingDown` and `linearDecayRoundingUp`, and the decay() functions for Input and Output, check endAmount < startAmount and call roudingDown/Up depending on that

test/reactors/V3DutchOrderReactor.t.sol
```diff
@@ -702,6 +703,54 @@ contract V3DutchOrderTest is PermitSignature, DeployPermit2, BaseReactorTest {
         assertEq(resolvedOrder.input.amount, 0);
     }
 
+    // 1000 - (100 * (1659087340-1659029740) / (65535)) = 912.1
+    // This is the input, which should round down to favor the swapper: 912
+    function testV3ResolveInputEndBlockAfterNow() public {
```
zhongeric, line 708: nit: no test for output positive / negative slope?

codyborn (commented):
I left [this comment](https://cantina.xyz/code/2848da7a-ebab-4d98-8957-40bc00dfeccb/comments#comment-5cef38d7-afe7-4290-86bc-ee3324d04cdb) in the audit. LMK what you think.

> Good to flag this. I think the difference this will make in practice will be quite small and wouldn't justify the additional complexity. The decay's will likely be < 50 bps across ~32 blocks. This means we'd be at most .99% off due to rounding (.495 bps).
> 
> It's also not clear that being too high or too low on the curve is actually beneficial for the user since we don't know what the real price is (and it's changing in real time). Imagine the actual price is 3. In one world we round up to 3 and get executed in block N. In another world we round down to 2 and hit it at block N+1 (at 4 instead).

src/lib/NonlinearDutchDecayLib.sol
```diff
@@ -46,7 +45,35 @@ library NonlinearDutchDecayLib {
         (uint16 startPoint, uint16 endPoint, int256 relStartAmount, int256 relEndAmount) =
             locateCurvePosition(curve, blockDelta);
         // get decay of only the relative amounts
-        int256 curveDelta = DutchDecayLib.linearDecay(startPoint, endPoint, blockDelta, relStartAmount, relEndAmount);
+        int256 curveDelta = v3LinearInputDecay(startPoint, endPoint, blockDelta, relStartAmount, relEndAmount);
+
+        return startAmount.boundedSub(curveDelta, minAmount, maxAmount);
+    }
+
+    function decayOutput(
```
codyborn, line 53: Can we combine these two functions to reduce code duplication?

codyborn: We can pass in a bool to differentiate input vs output

alanhwu: This was the approach I took in my first commit, but took @zhongeric's [suggestion](https://github.com/Uniswap/UniswapX/pull/305#discussion_r1786490670) to not use a boolean and separate out instead

codyborn (approved):

PR #312, opened 2024-10-17
Pass in decay function for input/output

codyborn:
Reduces code duplication by passing in the appropriate decay function


test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -11,17 +11,18 @@ import {Uint16Array, toUint256} from "../../src/types/Uint16Array.sol";
...
     function decayInput(
         NonlinearDutchDecay memory curve,
         uint256 startAmount,
         uint256 decayStartBlock,
-        uint256 minAmount,
         uint256 maxAmount
     ) public view {
-        NonlinearDutchDecayLib.decayInput(curve, startAmount, decayStartBlock, minAmount, maxAmount);
+        decayInput(curve, startAmount, decayStartBlock, maxAmount);
     }
```
alanhwu, line 26: Isn't this recursive?

test/lib/NonLinearDutchDecayLib.t.sol
```diff
@@ -31,12 +32,39 @@ contract MockNonlinearDutchDecayLibContract {
         uint256 minAmount,
         uint256 maxAmount
     ) public view {
-        NonlinearDutchDecayLib.decayOutput(curve, startAmount, decayStartBlock, minAmount, maxAmount);
+        decayOutput(curve, startAmount, decayStartBlock, minAmount, maxAmount);
```
alanhwu, line 35: Do we need `maxAmount` here if we can remove `minAmount` from `decayInput`?

test/lib/NonLinearDutchDecayLib.t.sol
codyborn: haha we actually don't need this altogether.. thanks for flagging!

PR #317, opened 2025-02-07
fix: add inline comments for internal revert tests

zhongeric:
the nightly version of forge raises errors for our tests which use vm.expectRevert on internal calls

https://book.getfoundry.sh/cheatcodes/expect-revert#error

### DIFF
--- a/test/lib/DutchDecayLib.t.sol
+++ b/test/lib/DutchDecayLib.t.sol
@@ -110,13 +110,13 @@ contract DutchDecayLibTest is Test {
     }
 
     /// forge-config: default.allow_internal_expect_revert = true
-    function testDutchDecayUnderflow() public {
+    function testDutchDownwardDecayOverflow() public {
         vm.expectRevert();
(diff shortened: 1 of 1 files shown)


test/lib/DutchDecayLib.t.sol
```diff
@@ -108,14 +109,19 @@ contract DutchDecayLibTest is Test {
         DutchDecayLib.decay(startAmount, endAmount, decayStartTime, decayEndTime);
     }
 
-    function testDutchDecayOverflow() public {
+    /// forge-config: default.allow_internal_expect_revert = true
+    function testDutchDecayUnderflow() public {
```
zhongeric, line 113: cc @codyborn I split this test up because per the foundry docs you should only have one internal revert being caught per test case. Just want to make sure the naming is correct here

alanhwu: These two cases are both overflows but one is for an upward decay (for exact_out) and the other for a normal downward decay. Will update the names to reflect that

alanhwu (approved):

PR #329, opened 2025-07-21
[WIP] Unified Reactor + DCA

ConjunctiveNormalForm:


src/interfaces/IDCARegistry.sol
```diff
@@ -0,0 +1,78 @@
...
+        address cosigner; // Trusted cosigner address
+        uint256 minFrequency; // Minimum time between chunks (seconds)
+        uint256 maxFrequency; // Maximum time between chunks (seconds)
+        uint256 minChunkSize; // Minimum input amount per chunk
+        uint256 maxChunkSize; // Maximum input amount per chunk
+        uint256 minOutputAmount; // Minimum output amount expected for this DCA execution
+        uint256 maxSlippage; // Max slippage in basis points (10000 = 100%)
+        uint256 deadline; // Intent expiration timestamp
+        uint256 nonce; // Replay protection nonce
+        bytes32 privateIntentHash; // Hash of private parameters (total amount, chunks, etc)
```
ConjunctiveNormalForm, line 29: User's commitment to the private dca params; since it's only the hash, smart contract will not be able to validate against them

maybe there can be an auxiliary dispute system where our TEE gets slashed if it creates orders that violate those constraints, and user choose to reveal the preimage?

src/interfaces/IValidationCallback.sol
```diff
@@ -8,5 +8,5 @@ interface IValidationCallback {
     /// @notice Called by the reactor for custom validation of an order. Will revert if validation fails
     /// @param filler The filler of the order
     /// @param resolvedOrder The resolved order to fill
-    function validate(address filler, ResolvedOrder calldata resolvedOrder) external view;
+    function validate(address filler, ResolvedOrder calldata resolvedOrder) external;
```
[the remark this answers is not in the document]
codyborn: Another option: Add a post-execution hook to update the state and leave this as a view call that validates using that state.

src/reactors/UnifiedReactor.sol
```diff
@@ -0,0 +1,85 @@
...
+    constructor(IPermit2 _permit2, address _protocolFeeOwner) BaseReactor(_permit2, _protocolFeeOwner) {}
+
+    /// @inheritdoc BaseReactor
+    function _resolve(SignedOrder calldata signedOrder)
+        internal
+        view
+        override
+        returns (ResolvedOrder memory resolvedOrder)
+    {
+        (address auctionResolver, bytes memory orderData) = abi.decode(signedOrder.order, (address, bytes));
```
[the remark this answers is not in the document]
codyborn: both are signed over by the swapper, right?

src/validation/DCARegistry.sol
```diff
@@ -0,0 +1,271 @@
...
+        if (order.info.additionalValidationData.length == 0) {
+            revert InvalidDCAParams();
+        }
+
+        bytes memory dcaData = order.info.additionalValidationData;
+
+        // Skip AllowanceTransfer prefix if present
+        if (dcaData.length > 0 && dcaData[0] == 0x01) {
+            // Remove the first byte (AllowanceTransfer flag) and decode the rest
+            bytes memory dcaValidationBytes = new bytes(dcaData.length - 1);
```
codyborn, line 56: This feels kinda janky but as you pointed out, we can get rid of it by adding the new field.

src/base/ReactorStructs.sol
```diff
@@ -47,6 +47,7 @@ struct ResolvedOrder {
     OutputToken[] outputs;
     bytes sig;
     bytes32 hash;
+    address auctionResolver;
```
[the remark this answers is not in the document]
zhongeric: I think this is at the wrong level of abstraction here because you _use_ an auctionResolver to create a ResolvedOrder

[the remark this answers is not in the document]
ConjunctiveNormalForm: yeah this specifically to make permit witness work for this PoC, because we need to correct typestring

maybe we can just add that to the ResolvedOrder struct too

src/reactors/UnifiedReactor.sol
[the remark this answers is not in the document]
ConjunctiveNormalForm: swapper sign the bytes that is `auction resolver address || specific-auction encoded order`

src/reactors/UnifiedReactor.sol
```diff
@@ -0,0 +1,85 @@
...
+        if (order.info.additionalValidationData.length > 0 && order.info.additionalValidationData[0] == 0x01) {
+            // Use AllowanceTransfer for DCA orders
+            IAllowanceTransfer.AllowanceTransferDetails memory details = IAllowanceTransfer.AllowanceTransferDetails({
+                from: order.info.swapper,
+                to: to,
+                amount: uint160(order.input.amount),
+                token: address(order.input.token)
+            });
+
+            IAllowanceTransfer(address(permit2)).transferFrom(details.from, details.to, details.amount, details.token);
```
marktoda, line 75: this is the root of my security concern:
- this now skips the `witness` check which guarantees onchain that the user actually approved the order details itself. now the reactor has no such validation at all and by this code path no authentication has been done at all (except maybe hidden internally in the resolver)

- secondly - full approving the reactor could be quite dangerous; in this case since `from` is just a details field - it seems possible to construct an order that pulls from someone else's hanging approval? I think this could be mitigated at various levels but is much more of a footgun than previously

src/validation/DCARegistry.sol
```diff
@@ -0,0 +1,271 @@
...
+        DCAValidationData memory validationData = abi.decode(dcaData, (DCAValidationData));
+        DCAIntent memory intent = validationData.intent;
+        DCAOrderCosignerData memory cosignerData = validationData.cosignerData;
+
+        // Calculate intent hash
+        bytes32 intentHash = hashDCAIntent(intent);
+
+        // Verify swapper signature if intent not already registered
+        if (!registeredIntents[intentHash]) {
+            _verifyIntentSignature(intent, validationData.signature, order.info.swapper);
```
codyborn, line 72: Is the only benefit of registeredIntents to save us from reverfying the signature?

src/reactors/UnifiedReactor.sol
zhongeric: Right it relies on external validation contracts rather than transfer with witness

src/reactors/UnifiedReactor.sol
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        _prepare(resolvedOrders);
+        IReactorCallbackV2(msg.sender).reactorCallback(resolvedOrders, callbackData);
+        _fill(resolvedOrders);
+    }
+
+    /// @notice Resolve a SignedOrder into a ResolvedOrderV2 using the auction resolver
+    function _resolve(SignedOrder calldata signedOrder) internal view returns (ResolvedOrderV2 memory resolvedOrder) {
+        (address auctionResolver, bytes memory orderData) = abi.decode(signedOrder.order, (address, bytes));
```
ConjunctiveNormalForm, line 100: alternatively, can add the auctionResolver field to the SignedOrder struct:
```
struct SignedOrder {
    bytes order;
    address auctionResolver;
    bytes sig;
}
```
}

src/base/ReactorStructs.sol
```diff
@@ -49,6 +70,16 @@ struct ResolvedOrder {
     bytes32 hash;
 }
 
+/// @dev generic concrete order that specifies exact tokens which need to be sent and received
+struct ResolvedOrderV2 {
+    OrderInfoV2 info;
+    InputToken input;
```
codyborn, line 76: For more flexibility, should we allow an array of input tokens? Could be used to batch liquidate all tokens to  ETH for example.

test/reactors/UnifiedReactor.t.sol
```diff
@@ -0,0 +1,746 @@
...
+        uint256 outputAmount = 1 ether;
+        uint256 deadline = block.timestamp + 1000;
+
+        // Seed both swapper and fillContract with enough tokens
+        tokenIn.mint(address(swapper), inputAmount);
+        tokenOut.mint(address(fillContract), outputAmount);
+        tokenIn.forceApprove(swapper, address(permit2), inputAmount);
+
+        // No need to set filler as valid - fillers are valid by default
+
```
codyborn, line 260: nit unnecessary whitespace

test/reactors/UnifiedReactor.t.sol
```diff
@@ -0,0 +1,746 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+import {Test, console} from "forge-std/Test.sol";
```
ericneil-sanc, line 4:
```suggestion
import {Test} from "forge-std/Test.sol";
```

PR #331, opened 2025-09-05
feat: UniswapX V4; DCA Hook

ConjunctiveNormalForm:
UniswapX now requires a hook for all orders, the minimum implementation of which should handle input token transfer from the swapper to the filler.

Added a base hook that all hooks should inherit from.


src/base/BaseHook.sol
```diff
@@ -0,0 +1,59 @@
...
+import {IPreExecutionHook} from "../interfaces/IPreExecutionHook.sol";
+import {ResolvedOrderV2} from "./ReactorStructs.sol";
+import {IPermit2} from "permit2/src/interfaces/IPermit2.sol";
+import {ISignatureTransfer} from "permit2/src/interfaces/ISignatureTransfer.sol";
+import {Permit2LibV2} from "../lib/Permit2LibV2.sol";
+import {IAuctionResolver} from "../interfaces/IAuctionResolver.sol";
+
+/// @notice Base hook contract that enforces token transfer logic
+/// @dev All hooks MUST inherit from this contract to ensure token transfers happen
+abstract contract BaseHook is IPreExecutionHook {
```
codyborn, line 13: Let's add a PostExecutionHook as well!

codyborn, line 13: Let's add a `PostExecutionHook` as well so users can take some action after the order is settled.

src/v4/base/BaseHook.sol
```diff
@@ -0,0 +1,58 @@
...
+
+    /// @notice Permit2 instance for signature verification and token transfers
+    IPermit2 public immutable PERMIT2;
+
+    constructor(IPermit2 _permit2) {
+        PERMIT2 = _permit2;
+    }
+
+    /// @inheritdoc IPreExecutionHook
+    function preExecutionHook(address filler, ResolvedOrder calldata resolvedOrder) external override {
```
codyborn, line 23: Should this only be callable by the Reactor?

src/v4/base/ReactorStructs.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+pragma solidity ^0.8.0;
+
+import {IReactor} from "../../interfaces/IReactor.sol";
+import {IPreExecutionHook, IPostExecutionHook} from "../interfaces/IHook.sol";
+import {InputToken, OutputToken} from "../../base/ReactorStructs.sol";
+
+/// @dev generic order information
+///  should be included as the first field in any concrete order types
+struct OrderInfo {
```
codyborn, line 10: Why is the `auctionResolver` not included here?

ConjunctiveNormalForm: maybe makes sense to have it in OrderInfo too, so that we can verify it against user signature

src/v4/base/BaseHook.sol
ConjunctiveNormalForm: not sure I follow this comment 
>Since the Reactor uses nonReentrant, there's no need to add it here, right?

But I think an `onlyReactor` modifier makes sense?

test/v4/Reactor.t.sol
```diff
@@ -0,0 +1,896 @@
...
+        uint256 feeAmount = uint256(outputAmount) * feeBps / 10000;
+        assertEq(tokenIn.balanceOf(address(swapper)), swapperInputBalanceStart - inputAmount);
+        assertEq(tokenIn.balanceOf(address(fillContract)), fillContractInputBalanceStart + inputAmount);
+        assertEq(tokenOut.balanceOf(address(swapper)), swapperOutputBalanceStart + outputAmount);
+        assertEq(tokenOut.balanceOf(address(fillContract)), fillContractOutputBalanceStart - outputAmount - feeAmount);
+        assertEq(tokenOut.balanceOf(address(feeRecipient)), feeAmount);
+    }
+
+    /// @dev Basic execute test for native currency output
+    function test_executeNativeOutput() public {
```
codyborn, line 205: We should also add a test for native input since it's technically supported, just not enabled

codyborn (approved):
Looks good! Added some suggestions.

codyborn (approved):

PR #332, opened 2025-09-29
feat: UniswapX V4 DCA Hook

alanhwu:
DCA Hook for the UniswapX V4 Reactor

### DIFF
--- a/test/v4/hooks/dca/DCAHookHarness.sol
+++ b/test/v4/hooks/dca/DCAHookHarness.sol
@@ -35,7 +35,7 @@ contract DCAHookHarness is DCAHook {
         s.totalOutput = totalOutput;
     }
 
-    /// @notice Exposes the internal _validateAllocations function for testing
+    /// @notice Exposes the internal _validateAllocationStructure function for testing
     function validateAllocationStructure(OutputAllocation[] memory outputAllocations) external pure {
         _validateAllocationStructure(outputAllocations);
     }
--- a/test/v4/hooks/dca/DCAHook.t.sol
+++ b/test/v4/hooks/dca/DCAHook.t.sol
@@ -1018,7 +1018,7 @@ contract DCAHookTest is Test, DeployPermit2 {
     // ========================================
     // Output Allocations Tests
     // ========================================
-    function test_validateOutputAllocations_validSingleRecipient() public view {
+    function test_validateAllocationStructure_validSingleRecipient() public view {
         OutputAllocation[] memory allocations = new OutputAllocation[](1);
         allocations[0] = OutputAllocation({recipient: SWAPPER, basisPoints: 10000});
 
@@ -1026,7 +1026,7 @@ contract DCAHookTest is Test, DeployPermit2 {
         hook.validateAllocationStructure(allocations);
     }
 
-    function test_validateOutputAllocations_validWithFees() public view {
+    function test_validateAllocationStructure_validWithFees() public view {
         OutputAllocation[] memory allocations = new OutputAllocation[](2);
         allocations[0] = OutputAllocation({
             recipient: SWAPPER,
@@ -1041,14 +1041,14 @@ contract DCAHookTest is Test, DeployPermit2 {
         hook.validateAllocationStructure(allocations);
     }
 
-    function test_validateOutputAllocations_revertsEmptyArray() public {
+    function test_validateAllocationStructure_revertsEmptyArray() public {
         OutputAllocation[] memory allocations = new OutputAllocation[](0);
 
         vm.expectRevert(IDCAHook.EmptyAllocations.selector);
         hook.validateAllocationStructure(allocations);
     }
 
-    function test_validateOutputAllocations_revertsZeroAllocation() public {
+    function test_validateAllocationStructure_revertsZeroAllocation() public {
         OutputAllocation[] memory allocations = new OutputAllocation[](2);
         allocations[0] = OutputAllocation({recipient: SWAPPER, basisPoints: 10000});
         allocations[1] = OutputAllocation({
@@ -1060,7 +1060,7 @@ contract DCAHookTest is Test, DeployPermit2 {
         hook.validateAllocationStructure(allocations);
     }
 
-    function test_validateOutputAllocations_revertsBelow100Percent() public {
+    function test_validateAllocationStructure_revertsBelow100Percent() public {
         OutputAllocation[] memory allocations = new OutputAllocation[](2);
         allocations[0] = OutputAllocation({
             recipient: SWAPPER,
@@ -1075,7 +1075,7 @@ contract DCAHookTest is Test, DeployPermit2 {
         hook.validateAllocationStructure(allocations);
     }
 
-    function test_validateOutputAllocations_revertsExceedsDuringSum() public {
+    function test_validateAllocationStructure_revertsExceedsDuringSum() public {
         // Test that allocations exceeding 100% are caught at the end
         OutputAllocation[] memory allocations = new OutputAllocation[](3);
         allocations[0] = OutputAllocation({
@@ -1095,7 +1095,7 @@ contract DCAHookTest is Test, DeployPermit2 {
         hook.validateAllocationStructure(allocations);
     }
 
-    function test_validateOutputAllocations_manyRecipients() public view {
+    function test_validateAllocationStructure_manyRecipients() public view {
         OutputAllocation[] memory allocations = new OutputAllocation[](10);
 
         for (uint256 i = 0; i < 9; i++) {
@@ -1114,7 +1114,7 @@ contract DCAHookTest is Test, DeployPermit2 {
         hook.validateAllocationStructure(allocations);
     }
 
-    function testFuzz_validateOutputAllocations_validDistributions(uint8 numRecipients, uint256 seed) public view {
+    function testFuzz_validateAllocationStructure_validDistributions(uint8 numRecipients, uint256 seed) public view {
         vm.assume(numRecipients > 0 && numRecipients <= 10);
 
         OutputAllocation[] memory allocations = new OutputAllocation[](numRecipients);


src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+
+        uint256 totalBasisPoints;
+        for (uint256 i = 0; i < length;) {
+            uint256 basisPoints = outputAllocations[i].basisPoints;
+            if (basisPoints == 0) {
+                revert ZeroAllocation();
+            }
+
+            totalBasisPoints += basisPoints;
+            if (totalBasisPoints > 10000) {
```
ConjunctiveNormalForm, line 202: nit: 10000 to constant `BPS`

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+        // 7) Price floor check (1e18 scaling)
+        _validatePriceFloor(intent, cosignerData);
+
+        // 8) Validate outputs match allocations and meet requirements
+        _validateOutputDistribution(intent, cosignerData, resolvedOrder.outputs);
+    }
+
+    function _transferInputTokens(ResolvedOrder calldata order, address to) private {
+        // TODO: Implement token transfer logic with AllowanceTransfer
+    }
```
ConjunctiveNormalForm, line 115: latest commit in the mandatory-hook branch has a TokenTransferLib now

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+            bytes memory swapperSignature,
+            bytes32 privateIntentHash,
+            DCAOrderCosignerData memory cosignerData,
+            bytes memory cosignerSignature
+        ) = abi.decode(
+            resolvedOrder.info.preExecutionHookData, (DCAIntent, bytes, bytes32, DCAOrderCosignerData, bytes)
+        );
+
+        // 2) Compute intentId for state lookups
+        bytes32 intentId = keccak256(abi.encodePacked(intent.swapper, intent.nonce));
```
ConjunctiveNormalForm, line 50: so we settled on the intent nonce getting randomly generated each time yeah (and not track in our backend for collision)?

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+
+        // 4) Transfer input tokens
+        _transferInputTokens(resolvedOrder, filler);
+
+        // 5) Update execution state and get cumulative totals
+        (uint256 totalInputExecuted, uint256 totalOutputExecuted) =
+            _updateExecutionState(intentId, resolvedOrder.input.amount, resolvedOrder.outputs);
+
+        // 6) Emit executing chunk event
+        emit ExecutingChunk(
```
ConjunctiveNormalForm, line 65: nit: think `ChunkExecuted` is a bit better

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+    /// @dev intentId is computed as keccak256(abi.encodePacked(swapper, nonce))
+    mapping(bytes32 => DCAExecutionState) internal executionStates;
+
+    constructor(IPermit2 _permit2) {
+        permit2 = _permit2;
+        domainSeparator = DCALib.computeDomainSeparator(address(this));
+    }
+
+    /// @inheritdoc IPreExecutionHook
+    function preExecutionHook(address filler, ResolvedOrder calldata resolvedOrder) external override {
```
ConjunctiveNormalForm, line 37: we'll probably want this to be `onlyReactor`

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+        if (executionStates[intentId].cancelled) {
+            revert IntentAlreadyCancelled(intentId);
+        }
+        executionStates[intentId].cancelled = true;
+        emit IntentCancelled(intentId, swapper);
+    }
+
+    /// @notice Validates the swapper's EIP-712 signature over the DCA intent
+    /// @dev Verifies the signature using the provided private intent hash to reconstruct the full intent hash
+    /// @param intent The DCA intent to validate (with zeroed privateIntent field)
```
ConjunctiveNormalForm, line 140: not sure where is the best place to put this, but explaining why zeroing out the private intent will be nice

test/v4/hooks/dca/DCAHookHarness.sol
```diff
@@ -0,0 +1,52 @@
...
+import {DCAHook} from "../../../../src/v4/hooks/dca/DCAHook.sol";
+import {
+    DCAExecutionState,
+    OutputAllocation,
+    DCAIntent,
+    DCAOrderCosignerData
+} from "../../../../src/v4/hooks/dca/DCAStructs.sol";
+import {IPermit2} from "permit2/src/interfaces/IPermit2.sol";
+
+contract DCAHookHarness is DCAHook {
```
ConjunctiveNormalForm, line 13: why this?

src/v4/hooks/dca/DCAHook.sol
alanhwu: Yes

src/v4/hooks/dca/DCAHook.sol
alanhwu: I originally had it as that but changed it bc I thought it'd be misleading since it hasn't actually been executed at this point

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,479 @@
...
+        for (uint256 i = 0; i < outputs.length; i++) {
+            totalOutput += outputs[i].amount;
+        }
+
+        // Update state in memory
+        state.executedChunks++;
+        state.lastExecutionTime = block.timestamp;
+        state.totalInputExecuted += inputAmount;
+        state.totalOutput += totalOutput;
+        state.nextNonce++;
```
codyborn, line 411: Since `nextNonce` and `executedChunks` are always the same, can we just use one or the other?

src/v4/lib/TokenTransferLib.sol
```diff
@@ -23,4 +24,37 @@ library TokenTransferLib {
...
+    function allowanceTransferInputTokens(
+        IPermit2 permit2,
+        ResolvedOrder calldata order,
+        address to
+    ) internal {
+        permit2.transferFrom(
+            order.info.swapper,
+            to,
+            uint160(order.input.amount),
+            order.input.token
```
ericneil-sanc, line 57:
```suggestion
            address(order.input.token)
```
This needs to be cast to an address of else it doesn't compile

src/v4/hooks/dca/DCAHook.sol
ConjunctiveNormalForm: my rationale is that the tx is atomic, it either didn't happen, or it happened and chunk is already executed when you can observe it

alanhwu: Ok, reverted!

src/v4/hooks/dca/DCAStructs.sol
```diff
@@ -0,0 +1,59 @@
...
+    uint256 nonce;
+    uint256 execAmount; // Amount being executed (input for EXACT_IN, output for EXACT_OUT)
+    uint256 limitAmount; // Limit/bound (min output for EXACT_IN, max input for EXACT_OUT)
+    uint96 orderNonce; // Unique execution chunk identifier (uint96)
+}
+
+struct DCAExecutionState {
+    uint96 nextNonce; // Next valid nonce (packs with cancelled)
+    bool cancelled; // Cancellation flag
+    uint256 executedChunks; // Number of chunks executed
```
codyborn, line 49: You could make `executedChunks` a uint96 so it can fit into the same slot as `nextNonce` and `cancelled`.

src/v4/hooks/dca/DCAStructs.sol
```diff
@@ -0,0 +1,59 @@
...
+}
+
+struct OutputAllocation {
+    address recipient;
+    uint256 basisPoints; // Out of 10000 (100% = 10000)
+}
+
+struct DCAOrderCosignerData {
+    address swapper;
+    uint256 nonce;
```
codyborn, line 40: If you made `nonce` a uint96, we could pack it in the same slot as address (20 bytes).

src/v4/hooks/dca/DCAStructs.sol
```diff
@@ -0,0 +1,59 @@
...
+    uint256 totalAmount; // Total amount on the exact side (input for EXACT_IN, output for EXACT_OUT)
+    uint256 exactFrequency;
+    uint256 numChunks;
+    bytes32 salt;
+    bytes32[] oracleFeeds; // Array of possible oracle feed IDs
+}
+
+struct OutputAllocation {
+    address recipient;
+    uint256 basisPoints; // Out of 10000 (100% = 10000)
```
codyborn, line 35: `basisPoints` can be uint16 and packed with recipient (20 bytes)

src/v4/hooks/dca/DCAStructs.sol
codyborn: `execAmount` and `limitAmount` could be made `uint160` which is also the max size that permit2 transferFrom will accept.

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,505 @@
...
+            revert IntentExpired(block.timestamp, intent.deadline);
+        }
+        if (cosignerData.orderNonce != state.nextNonce) {
+            revert WrongChunkNonce(cosignerData.orderNonce, state.nextNonce);
+        }
+
+        // Period gating (enforce minPeriod/maxPeriod only after first execution)
+        if (state.executedChunks > 0) {
+            uint256 elapsed = block.timestamp - state.lastExecutionTime;
+            if (elapsed < intent.minPeriod) {
```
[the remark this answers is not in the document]
alanhwu: I'm using `maxPeriod=0` as a sentinel for the never expiring behavior. So if it's 0 there's no enforced maximum amount of time between chunks

`minPeriod=0` actually matches what it sounds like, the most extreme case being chunk N executing then chunk N+1 executing later in the same block

src/v4/hooks/dca/DCAHook.sol
```diff
@@ -0,0 +1,505 @@
...
+    /// @notice Validates chunk size is within the allowed bounds
+    /// @dev Checks that execAmount is within min/max chunk size for the given order type
+    /// @param intent The DCA intent containing chunk size constraints
+    /// @param cosignerData The cosigner data containing execution amounts
+    /// @param inputAmount The actual input amount from the resolved order
+    function _validateChunkSize(DCAIntent memory intent, DCAOrderCosignerData memory cosignerData, uint256 inputAmount)
+        internal
+        pure
+    {
+        if (intent.isExactIn) {
```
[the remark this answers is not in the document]
alanhwu: Simplified it!

For the `limitAmount` check, this function does the `inputAmount` checks against it. It had to be separate from the output amount check in `_validateOutputDistribution()` because we need to add the outputs up.

The `_validatePriceFloor` is a different purpose of checking against the `minPrice` set in the original swapper intent

src/v4/hooks/dca/DCAStructs.sol
alanhwu: Packed a different permutation after we decided to remove nextNonce

PR #335, opened 2025-10-17
feat: Hybrid auction resolver

ConjunctiveNormalForm:
Hybrid dutch + priority auction, using [Tribunal's implementation](https://github.com/Uniswap/Tribunal/blob/main/src/lib/PriceCurveLib.sol)

### DIFF
--- a/src/v4/lib/HybridOrderLib.sol
+++ b/src/v4/lib/HybridOrderLib.sol
@@ -61,7 +61,7 @@ library HybridOrderLib {
         "address cosigner,",
         "HybridInput input,",
         "HybridOutput[] outputs,",
-        "uint256 auctionTargetBlock,",
+        "uint256 auctionStartBlock,",
         "uint256 baselinePriorityFee,",
         "uint256 scalingFactor,",
         "uint256[] priceCurve)"


src/v4/lib/HybridOrderLib.sol
```diff
@@ -0,0 +1,211 @@
...
+    address recipient;
+}
+
+/// @notice Hybrid auction order combining Dutch decay and priority gas auctions
+struct HybridOrder {
+    OrderInfo info;
+    address cosigner;
+    HybridInput input;
+    HybridOutput[] outputs;
+    uint256 auctionStartBlock;
```
codyborn: Nit: I'm curious why you use `auctionStartBlock` for the order but `auctionTargetBlock` for the cosigner override. Perhaps we can have the cosigner's be `auctionStartBlockOverride` for consistency?

ConjunctiveNormalForm: following the same lingo we use in priority orders

src/v4/docs/HYBRID_AUCTION_RESOLVER.md
```diff
@@ -0,0 +1,156 @@
...
+## Core Mechanism
+
+We use Tribunal's implementation through importing the PriceCurveLib. Read more about the core mechanics in [their doc](https://github.com/Uniswap/Tribunal/blob/main/docs/PriceCurveLib-documentation.md).
+
+### Cosigner Overrides
+
+Optional cosigner can provide:
+
+- `auctionTargetBlock`: Override the auction start block
+- `supplementalPriceCurve`: Additional scaling factors to combine with base curve
```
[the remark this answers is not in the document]
ConjunctiveNormalForm: >Or would it be better to allow the cosigner to override maxAmount/minAmount as long as it's strictly an improvement?

I think we are effectively doing the same thing here? Also I don't see reason to diverge from Tribunal's implementation

src/v4/lib/HybridOrderLib.sol
```diff
@@ -0,0 +1,211 @@
...
+        return result;
+    }
+
+    /// @notice scale the input of a hybrid order for exact-in orders
+    /// @param input the input to scale
+    /// @param scalingFactor the scaling factor to use
+    /// @return input scaled down from maxAmount
+    function scale(HybridInput memory input, uint256 scalingFactor) internal pure returns (InputToken memory) {
+        return
+            InputToken({token: input.token, amount: input.maxAmount.mulWad(scalingFactor), maxAmount: input.maxAmount});
```
codyborn: We should add a check that `scalingFactor <= 1e18`

src/v4/resolvers/HybridAuctionResolver.sol
```diff
@@ -0,0 +1,124 @@
...
+                effectivePriceCurve =
+                    order.priceCurve.applyMemorySupplementalPriceCurve(order.cosignerData.supplementalPriceCurve);
+            }
+        }
+
+        if (auctionTargetBlock != 0 && block.number < auctionTargetBlock) {
+            revert InvalidAuctionBlock();
+        }
+
+        uint256 currentScalingFactor =
```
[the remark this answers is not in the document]
codyborn: Cool, should we add this as a constraint on the cosigner?

codyborn (approved):

PR #375, opened 2026-08-05
feat: deploy DutchV3 reactor + OrderQuoter to Ink (57073)

codyborn:
Rolls UniswapX DutchV3 out to **Ink (chainId 57073)**, following `playbook/NEW_CHAIN.md`. Contracts are deployed and source-verified; this PR records the deployment and the research behind it.

## Deployed

| Contract | Address | Explorer |
|---|---|---|
| `V3DutchOrderReactor` | `0x000000007A1C8e570011EeDF86A2A35593013cBA` | [verified](https://explorer.inkonchain.com/address/0x000000007A1C8e570011EeDF86A2A35593013cBA) |
| `OrderQuoter` | `0x00000000a3db63Df9078cBF3dF88B4CAdD5a7F58` | [verified](https://explorer.inkonchain.com/address/0x00000000a3db63Df9078cBF3dF88B4CAdD5a7F58) |

Both verified on Blockscout with solc `v0.8.30+commit.73712a01`, optimizer 1,000,000 runs. Post-deploy the reactor's `owner()` returns `0x2BAD8182…46CD` and `permit2()` the canonical Permit2 — checked on-chain, not inferred from the deploy log.

`protocolFeeOwner` is derived from `PoolManager.owner()` at `0x360e68fa…fb32`, verified live.

## No contract changes needed

Ink is a textbook OP-stack L2: 1s contiguous blocks, real dynamic EIP-1559 basefee (252–623 wei observed across ~20M blocks), native ETH with the canonical WETH predeploy, EVM-equivalent `CALLVALUE`/`BALANCE`, Permit2 + Arachnid CREATE2 both present. So no `BlockNumberish` branch, no `adjustmentPerGweiBaseFee = 0`, no API-boundary native-sentinel rejection, and no sub-second retry floor.

## The reactor shares Robinhood's address

Ink has the canonical `protocolFeeOwner` and the same post-4663 bytecode as Robinhood, so Robinhood's CREATE2 salt is reused verbatim and both chains converge on one reactor address. Verified before broadcast: that salt under Ink's initcode hash `0xa0bb393c…` derives `0x000000007A1C…3cBA` exactly.

The deployed *runtime* differs from Robinhood's by exactly 3 bytes (offsets `0xae5`, `0xaea`) — the `BlockNumberish` immutable, routing Robinhood through `ArbSys` and Ink through `block.number`. CREATE2 hashes the initcode, not the runtime, so the address is unaffected; that diff is positive confirmation the chain gating works on both chains.

This is a different address from the `0x000000005aF6…` shared by the 17 chains deployed before the 4663 branch, whose salt no longer derives from current `creationCode`.

## Salt-selection guidance, corrected

`salts.json`'s `canonical` block previously held only the **retired pre-4663 pair**, which made a fresh mine look mandatory for any new chain even when reuse was available. Fixed here:

- `canonical` now holds the current post-4663 pair and instructs reuse; the old pair is nested as `_retiredPreBlockNumberish4663` so the older chains' address stays explicable.
- The top-level comment states the decision procedure: read `PoolManager.owner()`, reuse the canonical pair if it matches, mine only if it differs.
- `NEW_CHAIN.md` §3.2 gains a **"reuse before you mine"** section with the same rule plus two traps: `mine-salt.sh` has no reuse check and will mine a redundant salt if invoked, and any `creationCode` change (e.g. adding a chainid to `BlockNumberish.sol`) retires the canonical pair and requires promoting the next first-deployed pair into it.

## Tooling

- 57073 RPC default added to both deploy wrappers.
- `deploy-quoter-multichain.sh`'s `CHAINS` list is now env-overridable (`CHAINS=${CHAINS:-…}`), mirroring `SALTS_JSON` in the sibling script. It was hardcoded, so scoping a run to one chain wasn't possible — running it unmodified would sweep every chain and broadcast anywhere the quoter was missing. That list is also missing 4663 and 5042 despite both being deployed.

## Follow-ups (not in this PR)

- **Phase 2** — `uniswapx-sdk` needs explicit `PERMIT2_MAPPING`, `REACTOR_ADDRESS_MAPPING[Dutch_V3]`, `UNISWAPX_ORDER_QUOTER_MAPPING`, and `EXCLUSIVE_FILLER_VALIDATION_MAPPING` entries for 57073 (Ink isn't in `NETWORKS_WITH_SAME_ADDRESS`). `@uniswap/sdk-core` already has full Ink coverage.
- **Phase 3** — trading-api has no `57073` reference yet, so `CHAIN_INFO_MAP` / `V3_BLOCK_LENGTH_BY_CHAIN` / `GAS_COMPARISON_MULTIPLIER_BY_CHAIN` are greenfield adds there. Classic routing already exists (`uniroute` has Ink fully wired), so `compareQuotes` has a real Classic leg.
- **Owner drift on four chains** — a full dry run surfaced that xlayer, soneium, and celo have rotated `PoolManager.owner()` to `0x044aAF33…B6b7`, and avalanche to `0x47eB0Cf1…B5a6`. Their committed salts are stale and need re-mining. The wrapper refuses to deploy (`SKIP-OWNER-DRIFT`), so nothing is broken today.
- **Missing RPC defaults** — neither wrapper has a default for 4663 or 5042, so both skip with `SKIP-NO-RPC-CONFIG`. `https://rpc.mainnet.chain.robinhood.com` is live and returns 4663; `robinhood.md` still says the mainnet URL is unpublished.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/scripts/deploy-v3-multichain.sh
+++ b/scripts/deploy-v3-multichain.sh
@@ -7,10 +7,13 @@
 # Per-chain config (chainId, name, RPC default, v4 PoolManager, owner, salt,
 # expectedReactor) lives in `playbook/chains/salts.json`. The owner is derived
 # from the v4 PoolManager.owner() per chain so the V3 reactor's
-# protocolFeeOwner matches the AMM's per-chain governance. The (salt,
-# expectedReactor) pair is mined together via create2crunch against
-# (PERMIT2, owner). Chains where owner == 0x2bad...46cd reuse Tempo's
-# canonical mined salt and converge on 0x000000005aF6...
+# protocolFeeOwner matches the AMM's per-chain governance. Chains whose owner
+# matches the `canonical` block in salts.json reuse that block's (salt,
+# expectedReactor) verbatim and converge on one address; chains with a
+# different owner get their own pair mined via scripts/mine-salt.sh. See that
+# file's top-level comment for the decision procedure -- the canonical pair is
+# bytecode-dependent and is NOT the 0x000000005aF6... used by the chains
+# deployed before chainid 4663 was added to src/base/BlockNumberish.sol.
 #
 # Per chain, runs preconditions before broadcasting (in this order):
 #   1. RPC reachable + correct chainId.
@@ -85,6 +88,7 @@ default_rpc() {
     42161)    echo "https://arb1.arbitrum.io/rpc" ;;
     42220)    echo "https://forno.celo.org" ;;
     43114)    echo "https://api.avax.network/ext/bc/C/rpc" ;;
+    57073)    echo "https://rpc-gel.inkonchain.com" ;;
     59144)    echo "https://rpc.linea.build" ;;
     81457)    echo "https://rpc.blast.io" ;;
     7777777)  echo "https://rpc.zora.energy" ;;
--- a/README.md
+++ b/README.md
@@ -215,6 +215,31 @@ wei), so `adjustmentPerGweiBaseFee` is set to 0 at order construction.
 Explorer links pending Arcscan mainnet; see
 [playbook/chains/arc.md](./playbook/chains/arc.md).
 
+## Ink (DutchV3)
+
+| Contract                      | Address                                      | Source                                                        |
+| ---                           | ---                                          | ---                                                           |
+| V3 Dutch Order Reactor        | [0x000000007A1C8e570011EeDF86A2A35593013cBA](https://explorer.inkonchain.com/address/0x000000007A1C8e570011EeDF86A2A35593013cBA)   | [V3DutchOrderReactor](./src/reactors/V3DutchOrderReactor.sol) |
+| OrderQuoter                   | [0x00000000a3db63Df9078cBF3dF88B4CAdD5a7F58](https://explorer.inkonchain.com/address/0x00000000a3db63Df9078cBF3dF88B4CAdD5a7F58)   | [OrderQuoter](./src/lens/OrderQuoter.sol)                     |
+| Permit2                       | 0x000000000022D473030F116dDEE9F6B43aC78BA3   | [Permit2](https://github.com/Uniswap/permit2)                 |
+
+Ink (chainId 57073) is a standard OP-stack L2 — 1s blocks, real dynamic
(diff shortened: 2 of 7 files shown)


scripts/deploy-v3-multichain.sh
```diff
@@ -85,6 +85,7 @@ default_rpc() {
     42161)    echo "https://arb1.arbitrum.io/rpc" ;;
     42220)    echo "https://forno.celo.org" ;;
     43114)    echo "https://api.avax.network/ext/bc/C/rpc" ;;
+    57073)    echo "https://rpc-gel.inkonchain.com" ;;
```
ericneil-sanc, line 91: the comment on line 13 is now misleading: it tells the reader that canonical-owner chains "converge on 0x000000005aF6…", but under the bytecode this repo currently builds, that convergence is impossible — the correct target is 0x000000007A1C…3cBA, exactly as the PR's updated salts.json and NEW_CHAIN.md now say.

ericneil-sanc (dismissed):

scripts/deploy-v3-multichain.sh
codyborn: Good catch — fixed in `094fb06e`. That header block still described the pre-4663 world while `salts.json` and `NEW_CHAIN.md` in this same PR describe the current one, which is the worst combination: the script a deployer actually reads would have pointed them at an address the current bytecode cannot produce.

Rewritten to describe the rule rather than a specific address, since the canonical pair is bytecode-dependent and will move again the next time a chainid is added to `BlockNumberish.sol`:

```
# Chains whose owner matches the `canonical` block in salts.json reuse that
# block'''s (salt, expectedReactor) verbatim and converge on one address; chains
# with a different owner get their own pair mined via scripts/mine-salt.sh. See
# that file'''s top-level comment for the decision procedure -- the canonical pair
# is bytecode-dependent and is NOT the 0x000000005aF6... used by the chains
# deployed before chainid 4663 was added to src/base/BlockNumberish.sol.
```

Checked the sibling scripts (`deploy-quoter-multichain.sh`, `mine-salt.sh`) for the same stale claim — neither has one.

ericneil-sanc (approved):

